PR #8, opened 2017-11-20
Package props

ethanbennett:
### What was wrong?

The `Package` object needs the properties `name` and `version` from the supplied lockfile, and its `__repr__` function should return a unique package representation using them. The representation can eventually include `Package.hash` as discussed to ensure uniqueness.

@pipermerriam @carver I created this branch from https://github.com/pipermerriam/ethereum-erc190/pull/7, which creates the `Package` object. So that one should probably be merged first, and then I can rebase to get rid of the doubled commits you'll see here now. Otherwise, this is ready for review. Any criticism of my Python style is **very** welcome.

### How was it fixed?

All of the above exist as specified here: https://github.com/pipermerriam/ethereum-erc190/issues/6

### DIFF
--- a/erc190/package.py
+++ b/erc190/package.py
@@ -14,7 +14,6 @@ class Package(object):
     def __init__(self, lockfile):
         """
         A lockfile can be:
-
         - filesystem path
         - parsed lockfile JSON
         - lockfile URI
@@ -32,3 +31,16 @@ class Package(object):
         except jsonValidationError:
             raise ValidationError
         self.package_data = package_data
+
+    def __repr__(self):
+        name = self.name
+        version = self.version
+        return "<Package {0}=={1}>".format(name, version)
+
+    @property
+    def name(self):
+        return self.package_data['package_name']
+
+    @property
+    def version(self):
+        return self.package_data['version']
--- a/tests/erc190/test_package.py
+++ b/tests/erc190/test_package.py
@@ -42,3 +42,18 @@ def test_ethpm_doesnt_instantiate_with_invalid_package(invalid_package):
 def test_ethpm_doesnt_instantiate_with_invalid_path(invalid_path):
     with pytest.raises(ValidationError):
         Package(invalid_path)
+
+
+def test_package_object_has_name_property(valid_package):
+    current_package = Package(valid_package)
+    assert current_package.name == "wallet"
+
+
+def test_package_object_has_version_property(valid_package):
+    current_package = Package(valid_package)
+    assert current_package.version == "1.0.0"
+
+
+def test_package_has_custom_str_repr(valid_package):
+    current_package = Package(valid_package)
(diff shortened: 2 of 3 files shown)


pipermerriam:
https://github.com/pipermerriam/ethereum-erc190/pull/7 is now merged if you want to rebase.

erc190/package.py
```diff
@@ -29,6 +29,23 @@ def __init__(self, lockfile):
...
+    def __repr__(self):
+        """
+        This can be expanded to include the
+        hashed content as discussed once
+        Package.hash exists
+        """
+
+        name = self.package_data['package_name']
+        version = self.package_data['version']
+        return "<Package " + name + "==" + version + ">"
```
pipermerriam: This is somewhat of a preference but, for consistency, can you change this to construct the string as follows.

```
return "<Package {0}=={1}>".format(name, version)
```

tests/erc190/test_package.py
```diff
@@ -39,6 +54,7 @@ def test_ethpm_doesnt_instantiate_with_invalid_package(invalid_package):
         "abcd",
     )
 )
+
```
pipermerriam: bonus whitespace! (can you remove)

erc190/package.py
```diff
@@ -29,6 +29,23 @@ def __init__(self, lockfile):
...
+
+    def name(self):
```
pipermerriam: Can you change these to have an `@property` decorator so that they can be accessed as `pkg.name` rather than `pkg.name()`

erc190/package.py
```diff
@@ -29,6 +29,23 @@ def __init__(self, lockfile):
...
         self.package_data = package_data
+
+    def __repr__(self):
+        """
+        This can be expanded to include the
+        hashed content as discussed once
+        Package.hash exists
+        """
+
+        name = self.package_data['package_name']
```
pipermerriam: Can you change this and the `version` lookup to use the properties (`self.name` and `self.version`)

erc190/package.py
ethanbennett: I get the impression it's pretty basic, but I wasn't aware of the `@property` decorator before this. Thanks!

tests/erc190/test_package.py
ethanbennett: Oops..."bonus" makes it sound positive, at least.

erc190/package.py
```diff
@@ -32,3 +31,16 @@ def __init__(self, lockfile):
         except jsonValidationError:
             raise ValidationError
         self.package_data = package_data
+
+    def __refr__(self):
```
pipermerriam: spelling: this should be `__repr__`

ethanbennett: I guess I was thinking `reference` instead of `representation`, oops. I just fixed it in the test too--that's why the CI failed.

PR #10, opened 2017-11-21
Add validation to catch build dependencies

### LINKED ISSUE #4: Parse build dependencies, stub out validation.
### What needs to be built?

We need to expand the validation of packages to look at the [`build_dependencies`](https://github.com/ethpm/ethpm-spec/blob/master/release-lockfile.spec.md#build-dependencies-build_dependencies) key.

For the initial implementation add the following functionality to the code which validates the package during instantiation of the `Package` object.

- *if* the `build_dependencies` key is present in the package and it is non-empty, raise a `NotImplementedError` indicating that validation and loading of build dependencies has not yet been implemented.
- This validation should occur after the *schema* validation step.

ethanbennett:
### What was wrong?

Since we can't yet manage a package's build dependencies, we need to throw an error when a Package object is instantiated with dependencies.

### How was it fixed?

Package's `__init__` function checks for build dependencies after it validates with the schema.

#### Cute Animal Picture

Close #4


erc190/exceptions.py
```diff
@@ -3,3 +3,16 @@ class ValidationError(Exception):
     Error to signal something does not pass a validation check.
     """
     pass
+
+
+class NotImplementedError(Exception):
```
pipermerriam: I think you're overriding the built-in [`NotImplementedError`](https://docs.python.org/3/library/exceptions.html#NotImplementedError) here.  You should just use the built-in

erc190/package.py
```diff
@@ -30,6 +31,10 @@ def __init__(self, lockfile):
             validate(package_data, schema_data)
         except jsonValidationError:
             raise ValidationError
+
+        if package_data['build_dependencies']:
```
pipermerriam: this will throw a `KeyError` if the `build_dependencies` key is not present (which is allowed).  Can you pull in the `owned` package/lockfile example as a test case for this.

pipermerriam:
Just merged the `erc190 -> ethpm` PR so this will need to be rebased.

pipermerriam:
@ethanbennett too lazy to re-add the comments.  Here's a screenshot of requested changes.

screen shot 2017-11-22 at 11 04 46 am

ethanbennett:
@pipermerriam Thanks for the screenshot, I'm addressing the other comments now 👍

erc190/exceptions.py
ethanbennett: I didn't realize this was a built-in exception--I removed the custom one now.

ethpm/package.py
```diff
@@ -30,6 +28,11 @@ def __init__(self, lockfile):
             validate(package_data, schema_data)
         except jsonValidationError:
             raise ValidationError
+
+        if 'build_dependencies' in package_data.keys():
+            if package_data['build_dependencies']:
+                raise NotImplementedError
```
ethanbennett: @pipermerriam nested `if`s smell funny to me. Is this how you would do it?

pipermerriam: I'd do something like:

```python
if package_data.get('build_dependencies'):
    ...
```

The `.get(...)` method on dictionaries return `None` by default if the key isn't present so this should work for all scenarios that need to be handled.

ethanbennett: Yeah, that's much better. Thanks!

PR #12, opened 2017-11-24
Package contract factory

### LINKED ISSUE #2: Exposing contract factories from the Package object
We need an API for fetching the various contract types that are available in a given package.

### What should be done

Make the following API modifications:

* Change the constructor of the `Package` object to accept an *optional* Web3 instance.

```python
class Package:
    def __init__(self, package_identifier, web3=None):
        if web3 is not None:
            self.web3 = web3
        ...
```

Add a new API: `Package.set_default_web3(web3)`

The `set_default_web3(...)` function can be used to set the default `web3` instance that the `Package` will use.

Add a new API: `Package.get_contract_type(name, web3=None)`

* The `name` should be validated to be a valid contract name matching the regex: `[a-zA-Z][-a-zA-Z0-9_]{0, 255}`
* If the `web3` argument is not `None` then it should be used as the `web3` instance used to generate the contract factory.  **otherwise** it should use the `web3` instance located at `self.web3`.  If `self.web3` is `None` then an exception should be raised.
* If a key by the given `name` is found in the `contract_types` of the package, it's data should be passed to the `web3.eth.contract` API to generate a contract factory class which should be used as the return value for `get_contract_type` function.

njgheorghita:
@pipermerriam I noticed it's `bytecode_runtime` in web3 but `runtime_bytecode` in the lockfile - not sure if it's worth changing to stay consistent across repos?
Want to double check that when calling `get_contract_type` with a web3 object - it should NOT override `self.web3`? I set it up in this repo so it does not override the default.
Any other feedback on this pr would be great.

### What was wrong?
Package object didn't expose a contract factory

### How was it fixed?
Update package object to expose contract factory api

#### Cute Animal Picture
giphy 29

closes #2


ethpm/package.py
```diff
@@ -11,27 +12,69 @@
...
+        schema_data = self._load_json('./ethpm/assets/release-lockfile.schema.v1.json')
+        package_data = self._load_json(package_identifier)
 
         try:
             validate(package_data, schema_data)
         except jsonValidationError:
             raise ValidationError
         self.package_data = package_data
 
+    def _load_json(self, path):
```
pipermerriam: This falls into the category of *utility* which means I'd relegate it to something under `ethpm.utils`.  My suggestion would be to split this out into two independent functions.

* one that is dedicated to loading schemas and knows *where* to find them so that you don't have to provide absolute paths (HINT: look in populus for the `ASSETS_DIR` as I'd suggest modeling it after that.)
* one that is dedicated to loading packages from disk as JSON files (which probably does exactly what this function does).

A good guideline is to check whether `self` is used in a method.  If it isn't the thing probably deserves to get split out into it's own function.  (same for `cls` with `classmethods`).

ethpm/package.py
```diff
@@ -11,27 +12,69 @@
...
+        Set the default Web3 instance.
+        """
+        self.web3 = web3
+
+    def get_contract_type(self, name, web3=None):
+        """
+        API to generate a contract factory class.
+        """
+        self._validate_contract_name(name)
+        current_web3 = None
```
pipermerriam: this line can be removed since the next `if/else` block always sets it.

ethpm/package.py
```diff
@@ -11,27 +12,69 @@
...
+
+        if web3 is not None:
+            current_web3 = web3
+        else:
+            current_web3 = self.web3
+        if current_web3 is None:
```
pipermerriam: I'd suggest adding a line of whitespace above this so that the two `if` blocks are visually separated.

ethpm/package.py
```diff
@@ -11,27 +12,69 @@
...
+            raise ValidationError
```
pipermerriam: This should get a message added so that when it happens it's clear what went wrong.

ethpm/package.py
```diff
@@ -11,27 +12,69 @@
...
+
+        if name in self.package_data['contract_types']:
+            contract = self.package_data['contract_types'][name]
+            contract_factory = current_web3.eth.contract(
+                abi=contract["abi"],
```
pipermerriam: the `abi` key is not guaranteed to be present.  Neither are `bytecode` or `runtime_bytecode`.  would recommend splitting this logic up as follows.

* a function `validate_minimal_contract_data_present(contract_data)` which validates that at least one of the `abi/bytecode/runtime_bytecode` keys are present.
* a function `generate_contract_factory_kwargs` which uses the `eth_utils.to_dict` decorator to build a dictionary of the keyword arguments to be passed to `web3.eth.contract`

Then you can do this:

```python
v

```python
validate_minimal_contract_data_present(contract_data)
contract_kwargs = generate_contract_factory_kwargs(contract_data)
contract_factory = w3.eth.contract(**contract_kwargs)
```

requirements-dev.txt
```diff
@@ -10,4 +10,5 @@ pytest>=3.2.1
 pytest-pythonpath>=0.3
 pytest-watch==4.*
 tox>=1.8.0
+web3>=3.16.0
```
pipermerriam: Can you go ahead and pin this to the v4 beta.  Check the Releases section of the web3 repo for the exact version.

tests/ethpm/test_package.py
```diff
@@ -1,34 +1,49 @@
...
+def invalid_package_id():
     return "./tests/ethpm/invalidLockfile.json"
 
 
-def test_ethpm_instantiates_with_valid_package(valid_package):
-    current_package = Package(valid_package)
-    assert current_package.package_identifier == valid_package
+@pytest.fixture()
+def web3():
+    return Web3(HTTPProvider('http://localhost:8545'))
```
pipermerriam: You *should* switch to use `web3.providers.eth_tester.EthereumTesterProvider` here.

pipermerriam:
Oh, and in the web3 repo we've been switching to use `w3` as the standard variable name for a web3 instance.  Thoughts on switching to this?

njgheorghita:
@pipermerriam Thanks a lot for the feedback, addressed your comments - let me know if there's anything else that needs changing.

ethpm/utils/contract.py
```diff
@@ -0,0 +1,46 @@
...
+    "abi", "bytecode", "runtime_bytecode"
+    """
+    if not any(key in contract_data.keys() for key in ("abi", "bytecode", "runtime_bytecode")):
+        raise ValidationError(
+            "Minimum required contract data (abi/bytecode/runtime_bytecode) not found."
+        )
+
+
+def validate_contract_name(name):
+    pattern = re.compile("^[a-zA-Z][-a-zA-Z0-9_]{0,255}$")
```
pipermerriam: It's small savings but moving this to a module level definition instead of re-defining it everytime this function runs will yield some albeit minor speed improvements.

pipermerriam: Went ahead and merged.  I figure we can do this cleanup separately.

PR #14, opened 2017-11-30
[WIP] Move pattern compilation to module level

njgheorghita:
@pipermerriam is this along the lines of what you were talking about in the last pr?

'It's small savings but moving this to a module level definition instead of re-defining it everytime this function runs will yield some albeit minor speed improvements.'

### DIFF
--- a/ethpm/utils/contract.py
+++ b/ethpm/utils/contract.py
@@ -22,9 +22,11 @@ def validate_minimal_contract_data_present(contract_data):
         )
 
 
+CONTRACT_NAME_REGEX = re.compile("^[a-zA-Z][-a-zA-Z0-9_]{0,255}$")
+
+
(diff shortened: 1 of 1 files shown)


ethpm/utils/contract.py
```diff
@@ -22,8 +22,10 @@ def validate_minimal_contract_data_present(contract_data):
         )
 
 
+pattern = re.compile("^[a-zA-Z][-a-zA-Z0-9_]{0,255}$")
```
pipermerriam: Yes, but lets use a name like `CONTRACT_NAME_REGEX`.

njgheorghita:
@pipermerriam Should be good to go - or close this pr since it's minor - and i'll include the change down the road.

PR #16, opened 2017-12-03
Validate deployments reference existing contracts.

### LINKED ISSUE #15: Validate that all deployments reference existing contract types
### What was wrong?

STUB

validate that within the `deployments` of a package, all of the contracts types are present in the `contract_types`.

njgheorghita:
### What was wrong?
No validation that all deployments reference existing contract types in a package.

@pipermerriam - I can't quite remember for what kind of lockfiles are in scope for this pr. But this validation is currently built for packages with deployments with a single uri.

Split the load & validation logic into separate functions when you instantiate a Package.

Also curious to know if you know a better refactor for `validate_package_deployments()` in `package_validation.py`

### How was it fixed?
Wrote the validation.

#### Cute Animal Picture

giphy 38

closes #15

### DIFF
--- a/ethpm/utils/package_validation.py
+++ b/ethpm/utils/package_validation.py
@@ -1,18 +1,24 @@
+import itertools
 import json
 import os
-
-from ethpm import ASSETS_DIR
-from ethpm.exceptions import ValidationError
 
 from jsonschema import (
     validate,
     ValidationError as jsonValidationError
 )
 
+from ethpm import ASSETS_DIR
+
+from ethpm.exceptions import ValidationError
+
+
 RELEASE_LOCKFILE_SCHEMA_PATH = os.path.join(ASSETS_DIR, 'release-lockfile.schema.v1.json')
 
 
-def _load_package_data(package_id):
+def load_package_data(package_id):
+    """
+    Load package json located in ASSETS_DIR.
+    """
     with open(os.path.join(ASSETS_DIR, package_id)) as package:
         return json.load(package)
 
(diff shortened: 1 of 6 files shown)


ethpm/utils/package_validation.py
```diff
@@ -22,20 +27,42 @@ def _load_schema_data():
...
+    """
+    Validate that a package's deployments contracts reference existing contract_types.
+    """
+    if set(("contract_types", "deployments")).issubset(package_data):
+        all_contract_types = list(package_data["contract_types"].keys())
+
+        deployments = [
+            deployment
+            for uri, deployment
+            in package_data["deployments"].items()
```
pipermerriam: this can be changed to 

```python
all_deployments = list(package_data["deployments"].values())
```

ethpm/utils/package_validation.py
```diff
@@ -22,20 +27,42 @@ def _load_schema_data():
...
+        ]
+        deployment_names = [
+            name
+            for name, value
+            in deployments[0].items()
```
pipermerriam: This will only check the 1st set of deployments.  I think you want to change to something like

```
all_deployment_names = set(itertools.chain.from_iterable(
    deployments.keys()
    for deployments
    in all_deployments
))
```

ethpm/utils/package_validation.py
```diff
@@ -22,20 +27,42 @@ def _load_schema_data():
...
+        ]
+
+        if not deployment_names <= all_contract_types:
```
pipermerriam: I'm not sure about using `<=` here.  Not very readable.  I think something like the following will work better.

```
missing_contract_types = set(all_deployment_names).difference(all_contract_type_names)
if missing_contract_types:
    ...  # raise fancy error.
```

ethpm/utils/package_validation.py
```diff
@@ -48,20 +49,17 @@ def validate_package_deployments(package_data):
...
-            in package_data["deployments"].items()
-        ]
-        deployment_names = [
-            name
-            for name, value
-            in deployments[0].items()
-        ]
-
-        if not deployment_names <= all_contract_types:
+        all_deployments = list(package_data["deployments"].items())
```
pipermerriam: Change `items` to `values()` and you can remove the `[1]` from the next comprehension.

ethpm/utils/package_validation.py
```diff
@@ -48,20 +49,17 @@ def validate_package_deployments(package_data):
...
+        all_deployment_names = set(itertools.chain.from_iterable(
+            deployment[1]
```
pipermerriam: Just noting to drop the `[1]` once you change the above to `values()`

pipermerriam:
Good to merge with those last changes.

njgheorghita:
@pipermerriam Thanks for the help, good to merge.

PR #17, opened 2017-12-06
Fetch deployed contract instances pt. 1

njgheorghita:
### What was wrong?
Need an API for retrieving deployed contract instances if a package has one.

### How was it fixed?
Started laying the groundwork for such an api

#### Cute Animal Picture
giphy-downsized-large 4

### DIFF
--- a/ethpm/deployments.py
+++ b/ethpm/deployments.py
@@ -1,0 +1,5 @@
+class Deployments:
+
+    def __init__(self, deployment_data, w3):
+        self.deployment_data = deployment_data
+        self.w3 = w3
--- a/ethpm/package.py
+++ b/ethpm/package.py
@@ -1,16 +1,22 @@
+from ethpm.deployments import Deployments
+
 from ethpm.exceptions import ValidationError
 
-from ethpm.utils.package_validation import (
-    load_package_data,
-    validate_package_against_schema,
-    validate_package_exists,
-    validate_package_deployments,
-)
 from ethpm.utils.contract import (
     generate_contract_factory_kwargs,
     validate_contract_name,
     validate_minimal_contract_data_present,
     validate_w3_instance,
+)
+from ethpm.utils.deployment_validation import (
+    validate_single_matching_uri,
+)
+from ethpm.utils.package_validation import (
+    load_package_data,
+    validate_package_against_schema,
+    validate_package_exists,
+    validate_package_deployments,
+    validate_deployments_are_present,
 )
 
 
@@ -73,3 +79,20 @@ class Package(object):
     @property
     def version(self):
         return self.package_data['version']
+
+    #
+    # Deployments
+    #
+
+    def get_deployments(self, w3):
+        """
+        API to retrieve instance of deployed contract dependency.
+        """
+        validate_w3_instance(w3)
+        validate_deployments_are_present(self.package_data)
+
+        all_blockchain_uris = self.package_data["deployments"].keys()
+        matching_uri = validate_single_matching_uri(all_blockchain_uris, w3)
+
+        deployment_data = self.package_data["deployments"][matching_uri]
+        return Deployments(deployment_data, w3)
--- a/ethpm/utils/chains.py
+++ b/ethpm/utils/chains.py
@@ -1,0 +1,105 @@
+import re
+
+from urllib import parse
+from cytoolz import curry
+
+from eth_utils import (
+    add_0x_prefix,
+    remove_0x_prefix,
+    is_integer,
+    encode_hex,
+)
+
+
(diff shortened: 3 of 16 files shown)


ethpm/package.py
```diff
@@ -73,3 +79,20 @@ def name(self):
     @property
     def version(self):
         return self.package_data['version']
+
+    #
+    # Deployments
+    #
+
+    def get_deployments(self, w3):
```
njgheorghita: Do we always require a `w3` instance here to get passed in? or should it be able to use the default if there is one?

ethpm/package.py
```diff
@@ -73,3 +79,20 @@ def name(self):
...
+        """
+        API to retrieve instance of deployed contract dependency.
+        """
+        validate_w3_instance(w3)
+        validate_deployments_are_present(self.package_data)
```
njgheorghita: Checks that `"deployments"` key is present and that it's not empty. Is that too much / breaking the `1 thing / line of code` rule?

ethpm/package.py
```diff
@@ -73,3 +79,20 @@ def name(self):
...
+
+        all_blockchain_uris = self.package_data["deployments"].keys()
+        matching_uri = validate_single_matching_uri(all_blockchain_uris, w3)
```
njgheorghita: Similar question here to the one just above.

ethpm/utils/chains.py
```diff
@@ -0,0 +1,105 @@
...
+    _, resource_type, _ = parse_BIP122_uri(value)
+    return resource_type == BLOCK
+
+
+@curry
+def check_if_chain_matches_chain_uri(web3, blockchain_uri):
+    chain_id, resource_type, resource_hash = parse_BIP122_uri(blockchain_uri)
+    genesis_block = web3.eth.getBlock('earliest')
+
+    if encode_hex(genesis_block['hash']) != chain_id:
```
njgheorghita: I added `encode_hex()` here and on line 68 to get it to work. Is that fine? When "stealing" utils from other repos, is it better to leave them as is - or adjust as necessary?

ethpm/package.py
pipermerriam: nah, that's fine.  functions are *wrappers* around chunks of functionality so they get a pass on that rule/guideline.

ethpm/utils/chains.py
pipermerriam: Adjust as necessary.  Most of the utils you're stealing from populus are going to be removed as this library will replace them.

ethpm/deployments.py
```diff
@@ -0,0 +1,5 @@
+class Deployments:
+
+    def __init__(self, deployment_data, w3):
+        self.deployment_data = deployment_data
```
pipermerriam: FYI, in the next pull request, we'll need to pass in the `contract_types` in some format.  I have some ideas on that that I'll post elsewhere.

ethpm/package.py
pipermerriam: I think that it's ok to make it optional `get_deployments(self, w3=None)` and to default to the class `w3` instance if no instance is passed in.

ethpm/utils/chains.py
```diff
@@ -0,0 +1,105 @@
...
+        remove_0x_prefix(chain_id),
+        "{0}/{1}".format(resource_type, remove_0x_prefix(resource_identifier)),
+        '',
+        '',
+    ])
+
+
+def create_block_uri(chain_id, block_identifier):
+    if is_integer(block_identifier):
+        return create_BIP122_uri(chain_id, 'block', str(block_identifier))
```
pipermerriam: I think that this code path can be removed.  We won't be using block uris that use block numbers.

tests/conftest.py
```diff
@@ -33,16 +54,89 @@
...
+    }
+    f = tmpdir.join("lockfile.json")
+    f.write(json.dumps(lockfile))
+    return str(f)
+
+
+@pytest.fixture
+def lockfile_with_no_matching_deployments(w3, tmpdir):
+    w3.testing.mine(5)
+    incorrect_chain_id = get_chain_id(w3)[:-1] + b'X'
```
pipermerriam: It'd be better to just use a totally bogus hash rather than modifying the existing one.

```python
incorrect_chain_id = encode_hex(b'\x00' * 31 + b'\x01')  # or just use the longhand string and drop the `encode_hex` call.
```

pipermerriam:
I went ahead and merged.  There was some minor cleanup that maybe you can roll into the next PR?

njgheorghita:
@pipermerriam Sounds good, I'll make sure they get included.

PR #19, opened 2017-12-08
API for initializing Package objects

### LINKED ISSUE #18: API for initializing Package objects
### What was wrong?

Right now the `Package.__init__` *expects* to be given a path to a lockfile.  Ideally, we'd like to be able to instantiate `Package` objects with any of:

* Filesystem path
* Actual lockfile data (the parsed JSON data)
* lockfile URI
* other things we haven't thought of yet.

### How can it be fixed?

We *could* make the `Package.__init__` method accept any of the formats and do content negotiation there.  However this approach is likely to result in complex initialization logic that is hard to test and isolate.

Instead lets do the following.

- Modify the `Package` object to **only** accept the package data.
- Add the following `classmethod` methods on the `Package` object:
  - `Package.from_file`
  - `Package.from_uri`

Both of the `from_*` methods will retrieve the lockfile JSON from their respective sources and then initialize a new `Package` object.

ethanbennett:
### What was wrong?

The package object currently expects a filepath to access its lockfile.

### How was it fixed?

This PR updates `Package` to take a lockfile object, and it provides a class method for converting lockfiles provided as paths into JSON objects (so they can be used to initialize Packages).

#### Cute Animal Picture

Cute animal picture

Closes #18

### DIFF
--- a/ethpm/package.py
+++ b/ethpm/package.py
@@ -1,3 +1,5 @@
+import json
+
 from ethpm.deployments import Deployments
 
 from ethpm.exceptions import ValidationError
@@ -13,28 +15,44 @@ from ethpm.utils.deployment_validation import (
 )
 from ethpm.utils.package_validation import (
     check_for_build_dependencies,
-    load_package_data,
     validate_package_against_schema,
-    validate_package_exists,
     validate_package_deployments,
     validate_deployments_are_present,
 )
 
 
+def _load_package_data_from_file(file_obj):
+    """
+    Utility function to load package objects
+    from file objects passed to Package.from_file
+    """
+    try:
+        package_data = json.load(file_obj)
+    except json.JSONDecodeError as err:
+        raise json.JSONDecodeError(
+            "Failed to load package data. File is not a valid JSON document.",
+            err.doc,
+            err.pos,
+        )
+
+    return package_data
+
+
 class Package(object):
 
-    def __init__(self, package_id, w3=None):
+    def __init__(self, package_data, w3=None):
         """
-        A lockfile can be:
-        - filesystem path
-        - parsed lockfile JSON
-        - lockfile URI
+        A lockfile must be constructed with
+        parsed lockfile JSON.
         """
         self.w3 = w3
-        self.package_id = package_id
 
-        validate_package_exists(package_id)
-        package_data = load_package_data(package_id)
+        if not isinstance(package_data, dict):
+            raise TypeError(
+                "Package object must be initialized with a dictionary. "
+                "Got {0}".format(type(package_data))
+            )
+
         validate_package_against_schema(package_data)
         validate_package_deployments(package_data)
         check_for_build_dependencies(package_data)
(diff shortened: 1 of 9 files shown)


ethpm/package.py
```diff
@@ -72,6 +86,18 @@ def __repr__(self):
...
 
+    @classmethod
+    def from_file(cls, file_path_or_obj):
+        if isinstance(file_path_or_obj, str):
+            with open(file_path_or_obj) as file_obj:
+                package_data = _load_package_data_from_file(file_obj)
+        elif hasattr(file_path_or_obj, 'read') and callable(file_path_or_obj.read):
+            package_data = _load_package_data_from_file(file_path_or_obj)
+        else:
+            assert False
```
pipermerriam: I believe that this case should raise a `TypeError` with a message something like:

> `The `Package.from_file` method takes either a filesystem path or a file-like object.  Got {0} instead`

Fill `{0}` in with `type(file_path_or_obj)`

tests/test_package_initialization.py
```diff
@@ -0,0 +1,87 @@
...
+
+@pytest.fixture(params=[True, False])
+def non_json_lockfile(tmpdir, request):
+    file = tmpdir.mkdir('invalid').join('lockfile.json')
+    file.write('This is invalid json')
+    if request.param:
+        yield str(file)
+    else:
+        with open(str(file)) as file_obj:
+            yield file_obj
```
pipermerriam: Here is a pattern for making this generic.

```python

def _generate_fixture_param(tmpfile, as_file_path, file_content):
    temp_lockfile = tmpdir.mkdir('invalid').join('lockfile.json')
    temp_lockfile.write(file_content)
    if as_file_path:
        return str(temp_lockfile)
    else:
        with open(str(file)) as file_obj:
            return file_obj


@pytest.fixture(params=[True, False])
def the_first_fixture(tmpfile, request):
    return _generate_fixture_param(tmpfile, request.param, 'This is invalid json')

@pytest.fixture(params=[True, False])
def the_second_fixture(tmpfile, request):
    return _generate_fixture_param(tmpfile, request.param, '{"something": "else"}')
```

ethanbennett: Awesome. I'm hoping to get some more done here this afternoon.

ethpm/package.py
```diff
@@ -12,28 +14,42 @@
...
 
+def _load_package_data_from_file(file_obj):
+    try:
+        package_data = json.load(file_obj)
+    except json.JSONDecodeError as err:
+        raise json.JSONDecodeError(
+            "Failed to load package data. File is not a valid JSON document.",
+            err.doc,
+            err.pos
+            )
```
pipermerriam: This is valid python but we typically would dedent this one more level to be inline with the opening line.  Also, we like trailing commas in python land.

tests/conftest.py
```diff
@@ -53,13 +53,26 @@ def w3():
...
+
+
 @pytest.fixture
 def lockfile_with_no_deployments(tmpdir):
     lockfile = copy.deepcopy(LOCKFILE)
     lockfile.pop("deployments")
     f = tmpdir.join("lockfile.json")
     f.write(json.dumps(lockfile))
-    return str(f)
+    return json.load(f)
```
pipermerriam: Note what this function is currently doing.  Taking the json, writing it to a file, then returning the json parsed from the file.  You can shorten this whole function (and the subsequent fixtures) to just `return` the lockfile data and remove all of the *file* nonsense.

ethanbennett: Very good point. I did this from the perspective of fixing the tests and hadn't zoomed out enough to see the obvious.

tests/test_package_initialization.py
ethanbennett: @pipermerriam Did you mean `tmpdir` where you wrote `tmpfile`? I initially read it that way but now I'm hitting some bugs, so I want to be sure they're not coming from a bad assumption.

ethpm/package.py
ethanbennett: Got it 👍

tests/test_package_initialization.py
pipermerriam: yes, `tmpdir`, my bad.

ethanbennett: Awesome. I think I figured out what's wrong, hoping to get this ready for review this evening.

pipermerriam:
Looks like ff4a441 fixed it.

ethanbennett:
I'll still get that integrated with other valid tests, and I'm going to try to consolidate some of these fixtures too.

PR #20, opened 2017-12-12
Fetch deployed contract instances pt 2

### LINKED ISSUE #13: API for getting instances of a deployed contract instance
### What was wrong?

Packages can contain the addresses of deployed contract instances.  We need an API for retrieving these.

### How can it be fixed?

Since packages can contain deployments across multiple chains we'll need an API that allows us to access these.  I propose the following two APIs

* `Package.deployments`
* `Package.get_deployments(w3)`

The `Package.deployments` API would use the default web3 instance that the `Package` is currently configured with.  If the `Package` has no web3 instance accessing this property should throw an exception indicating that a an instance of web3 is reuired.

The `Package.get_deployments(w3)` allows a user to specify a web3 instance.

Both of these APIs should return a `Deployments` object which exposes a subset of the normal dictionary interface.

* `Deployments.get('MyContract')`
* `Deployments['MyContract']`
* `'MyContract' in Deployments`
* `Deployments.keys()`
* `Deployments.items()`
* `Deployments.values()`

TODO: figure out what to do about validation and what that API looks like.  Does this issue need to stub out any of that functionality.

njgheorghita:
### What was wrong?
Deployments object needed basic functionality (pt 2 of API for getting instances of a deployed contract instance).

### How was it fixed?
Completed checklist from comment in #13 

#### Cute Animal Picture
giphy-downsized-large 5

Closes #13

### DIFF
--- a/ethpm/deployments.py
+++ b/ethpm/deployments.py
@@ -10,9 +10,9 @@ class Deployments:
     Deployment object to access instances of
     deployed contracts belonging to a package.
     """
-    def __init__(self, deployment_data, contract_types, w3):
+    def __init__(self, deployment_data, contract_factories, w3):
         self.deployment_data = deployment_data
-        self.contract_types = contract_types
+        self.contract_factories = contract_factories
         self.w3 = w3
 
     def __getitem__(self, key):
@@ -26,11 +26,19 @@ class Deployments:
         return self.deployment_data.get(key)
 
     def items(self):
-        item_dict = {name: self.get(name) for name in self.deployment_data}
+        item_dict = {
+            name: self.get(name)
+            for name
+            in self.deployment_data
+        }
         return item_dict.items()
 
     def values(self):
-        values = [self.get(name) for name in self.deployment_data]
+        values = [
+            self.get(name)
+            for name
+            in self.deployment_data
+        ]
         return values
 
     def get_contract_instance(self, contract_name):
@@ -47,5 +55,5 @@ class Deployments:
         if name not in self.deployment_data:
             raise KeyError("Contract name not found in deployment data")
 
-        if name not in self.contract_types:
-            raise ValidationError("Contract name not found in contract_types")
+        if name not in self.contract_factories:
+            raise ValidationError("Contract name not found in contract_factories.")
--- a/ethpm/deployments.py
+++ b/ethpm/deployments.py
@@ -1,5 +1,51 @@
-class Deployments:
+from ethpm.exceptions import ValidationError
 
-    def __init__(self, deployment_data, w3):
+from ethpm.utils.contract import (
+    validate_contract_name,
+)
+
+
+class Deployments:
+    """
+    Deployment object to access instances of
+    deployed contracts belonging to a package.
+    """
+    def __init__(self, deployment_data, contract_types, w3):
         self.deployment_data = deployment_data
+        self.contract_types = contract_types
         self.w3 = w3
+
+    def __getitem__(self, key):
+        return self.get(key)
+
+    def __contains__(self, key):
+        return key in self.deployment_data
+
+    def get(self, key):
+        self._validate_name_and_references(key)
+        return self.deployment_data.get(key)
+
+    def items(self):
+        item_dict = {name: self.get(name) for name in self.deployment_data}
+        return item_dict.items()
+
+    def values(self):
+        values = [self.get(name) for name in self.deployment_data]
+        return values
+
+    def get_contract_instance(self, contract_name):
+        """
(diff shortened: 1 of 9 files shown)


ethpm/deployments.py
```diff
@@ -1,5 +1,50 @@
...
+
+    def get_contract_instance(self, contract_name):
+        """
+        Fetches a contract instance belonging to deployment
+        after validating contract name.
+        """
+        self._validate_name_and_references(contract_name)
+        raise NotImplementedError("All checks passed, but get_contract_instance API not complete.")
+
+    def _validate_name_and_references(self, name):
```
njgheorghita: should this be pulled out into a util?

ethpm/deployments.py
```diff
@@ -1,5 +1,50 @@
...
+)
+
 
-    def __init__(self, deployment_data, w3):
+class Deployments:
+    """
+    Deployment object to access instances of
+    deployed contracts belonging to a package.
+    """
+    def __init__(self, deployment_data, contract_types, w3):
```
njgheorghita: Wasn't sure the best way to handle the contract_types, so for now just passing in the `"contract_types".keys()` from `package_data` for validation - if there's a better way to handle it, I can make the changes.

ethpm/deployments.py
pipermerriam: No, but what I would do this.

* Add a new `.get(contract_name)` method and inline this validation in that method.
* Rewrite *all* of the `values()`, `items()`, and `__getitem__()` methods to use `self.get(contract_name)` internally so that there is a single point where validation happens and all other APIs make use of that single point of entry.

njgheorghita:
@pipermerriam addressed the first comment. I think I have a pretty good understanding of what you're looking for in the second comment. Basically pass in contract factory objects (as kwargs?) when instantiating Deployments? Maybe? 

Is that something you want implemented in this or a later pr?

pipermerriam:
> Is that something you want implemented in this or a later pr?

Fine to do it as a *part 3* to this set of changes if you want to implement all the failure cases first and then do the success case as a 3rd PR.

ethpm/package.py
```diff
@@ -84,15 +84,24 @@ def version(self):
...
         all_blockchain_uris = self.package_data["deployments"].keys()
         matching_uri = validate_single_matching_uri(all_blockchain_uris, w3)
 
         deployment_data = self.package_data["deployments"][matching_uri]
-        return Deployments(deployment_data, w3)
+
+        all_contract_factories = {}
+        all_contract_names = list(self.package_data["contract_types"].keys())
+        for name in all_contract_names:
+            all_contract_factories[name] = self.get_contract_type(name, w3)
```
pipermerriam: You can use a dictionary comprehension

```python
all_contract_factories = {
    name: self.get_contract_type(name, w3) 
    for name in 
    self.package_data["contract_types"].keys()
}
```

Note that `package_data` isn't required to have a `contract_types` key.  Also, it may be prudent to not put *all* contract types in, but rather to only put in the types necessary for the deployments.

```python
deployments = self.package_data["deployments"][matching_uri]
all_contract_factories = {
    deployment_data['contract_type']: self.get_contract_type(deployment_data['contract_type'], w3) 
    for deployment_data
    in deployments.values()
}
```

This way you are passing in the absolute minimal information necessary for the `Deployments` object to do it's magic.

tests/ethpm/test_get_deployments.py
```diff
@@ -38,7 +49,8 @@ def test_get_deployments_with_multiple_matches_raises_exception(w3, lockfile_wit
...
 
 
-def test_get_deployments_with_a_match_returns_deployments(w3, lockfile_with_matching_deployment):
-    package = Package(lockfile_with_matching_deployment)
-    deployment = package.get_deployments(w3)
+def test_get_deployments_with_a_match_returns_deployments(w3, matching_package):
+    deployment = matching_package.get_deployments(w3)
+    contract_factory = deployment.contract_factories["SafeMathLib"]
     assert isinstance(deployment, Deployments)
+    assert getattr(contract_factory, '__module__') == web3.contract.__name__
```
njgheorghita: I was trying to come up with a better way to test the contract_factory is a contract_factory, but I couldn't get isinstance() to cooperate. Is this fine / is there a better way?

pipermerriam: Hrm, maybe it'd be better to `assert "SafeMathLib" in deployments` to simply test for presence.

njgheorghita:
@pipermerriam Thanks! I think that's it, let me know if I missed something.

PR #22, opened 2017-12-21
WIP Nick/fixture tests

njgheorghita:
Hey @pipermerriam, I took a shot at writing the fixture tests. I jacked a lot of the code / design from py-evm. Not sure if that was the best way to approach (adding unnecessary complexity)? Should I go ahead and create a json fixture test file for all the different kinds of packages? What other kinds of things belong in the fixture file? Right now it's really only testing that the `Package` object has certain attributes. Feedback on what's good / bad / needs changing would be really helpful, thanks! 

### What was wrong?
Project needs standardized json fixture tests to allow other py-ethpm implementations to develop against the same spec. 

### How was it fixed?
Started laying the groundwork for fixture tests. 

#### Cute Animal Picture
giphy 39


pipermerriam:
I merged ethan's pr which caused a merge conflict.  Also, his PR changes the API for initializing `Package` objects which I'm betting is going to cause some failures here.  No rush (this PR looks really good btw)

PR #23, opened 2018-01-04
Write json fixture test groundwork

### LINKED ISSUE #9: JSON file based test suite
### What is wrong?

It would be good if we could standardize our tests into a set of portable JSON documents which describe the *inputs* and *expected results* for things like loading and validating packages and their internal data.  This would serve as a framework for other languages which implement a similar library to ensure that everyone is developing against the same spec.

### How can it be fixed?

Similar to how the `ethereum/tests` repository contains JSON based tests for EVM implementations, we should develop the same.  We can start with them living in this repository and then move them to their own repository in due time.

njgheorghita:
Hey @pipermerriam, I took a shot at writing the fixture tests. I jacked a lot of the code / design from py-evm. Not sure if that was the best way to approach (adding unnecessary complexity)? Should I go ahead and create a json fixture test file for all the different kinds of packages? What other kinds of things belong in the fixture file? Right now it's really only testing that the `Package` object has certain attributes. Feedback on what's good / bad / needs changing would be really helpful, thanks! 

### What was wrong?
Project needs standardized json fixture tests to allow other py-ethpm implementations to develop against the same spec. 

### How was it fixed?
Started laying the groundwork for fixture tests. 

#### Cute Animal Picture
giphy 39

closes #9

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ lib
 lib64
 venv*
 .DS_Store
+.pytest_cache/
 
 # Installer logs
 pip-log.txt
--- a/ethpm/__init__.py
+++ b/ethpm/__init__.py
@@ -11,6 +11,6 @@ if sys.version_info.major < 3:
 
 
 BASE_DIR = os.path.dirname(__file__)
-ASSETS_DIR: str = os.path.join(BASE_DIR, 'assets')
+ASSETS_DIR = os.path.join(BASE_DIR, 'assets')  # type: str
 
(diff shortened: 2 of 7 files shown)


njgheorghita:
Hey @pipermerriam ran into some nasty git issues with the old branch, but this is an exact duplicate, adjusted for ethan's pr. If there's anything else that needs to be added, holler.

pipermerriam:
will take a look at it tomorrow.

PR #24, opened 2018-04-30
Convert ci from travis to circle

njgheorghita:
### What was wrong?
Repo needed updating from travis ci to circle ci

### How was it fixed?
Remove `.travis.yml`, create `.circleci/config.yml`

#### Cute Animal Picture
image


njgheorghita:
@pipermerriam I'm having some trouble figuring out what's going wrong with my `config.yml`, any thoughts? are the `checkout fixtures submodule` and `merge pull request base` commands from https://github.com/ethereum/py-evm/issues/604 necessary?

njgheorghita:
hmmmm - i managed to get the tests passing, but if you wouldn't mind giving the config file a look over to make sure nothing stands out as wrong, i'd feel a lot better about it

pipermerriam:
Looks like you're missing a large part.

https://github.com/ethereum/eth-abi/blob/master/.circleci/config.yml#L7-L24

That section is what verifies that the branch not only passed on its own, but also that the resulting *merge* passes as well.  You *should* be able to largely copy/paste that section.

You also *might* go ahead and add a `pypy` run too.  Here is the config for that from `eth-utils`: https://github.com/ethereum/eth-utils/blob/master/.circleci/config.yml#L61-L66

pipermerriam:
If pypy3 doesn't pass right away just drop it and we can re-add later.

njgheorghita:
yup, wasn't able to get pypy3 working, but everything else should be good now

pipermerriam:
Try pinning `rlp<1.0` for now.

tests/ethpm/test_package.py
```diff
@@ -67,4 +67,3 @@ def test_package_object_has_version_property(valid_lockfile):
 def test_package_has_custom_str_repr(valid_lockfile):
     current_package = Package(valid_lockfile)
     assert current_package.__repr__() == "<Package wallet==1.0.0>"
-
```
davesque: @njgheorghita It's a bit hard to tell from the diff what this change was.  If the newline at the end of the file was removed, in the future it would be to avoid that kind of a whitespace change.  Practically speaking, it doesn't really matter if a newline is at the end of a file but a lot of editors add it anyway to be POSIX compliant.  It's best just to leave them in so there aren't a lot of commits that undo and redo the same edits.

njgheorghita: @davesque Cool, good to know. If I remember correctly, flake8 was yelling at me b/c of the newline at end of file. But I was confused as to why the newline was there in the first place (if flake8 yelled now, it should have yelled earlier??)

davesque: @njgheorghita Hmm...I would wager that flake8 was complaining about multiple newlines at the end of the file.  Here's a more explicit example of what I'm talking about:
```
$ printf 'foo = "bar"' > test.py
$ flake8 test.py
test.py:1:12: W292 no newline at end of file

$ printf 'foo = "bar"\n' > test.py
$ flake8 test.py    # Generates no errors because file ends with newline

$ printf 'foo = "bar"\n\n' > test.py
$ flake8 test.py
test.py:2:1: W391 blank line at end of file
```

njgheorghita: Ahh ok got it, thanks!

PR #25, opened 2018-05-02
Rename lockfile to package throughout repo

njgheorghita:
### What was wrong?
"Lockfile" is a confusing name, so renamed all occurrences of "Lockfile" to "Package"

#### Cute Animal Picture
image


ethpm/assets/release-package.schema.v2.json
```diff
@@ -0,0 +1,279 @@
...
+    "version"
+  ],
+  "version": "2",
+  "additionalProperties": false,
+  "properties": {
+    "package_version": {
+      "type": "string",
+      "title": "Package Version",
+      "default": "2",
+      "enum": ["2"]
```
njgheorghita: My idea was that this will eventually be the schema for V2. But I'm worried that "package_version" might get confusing with the "version" field below? Is there a better name we can use?

cburgdorf:
>"Lockfile" is a confusing name, so renamed all occurrences of "Lockfile" to "Package"

Package managers often *do* have a concept of lockfiles so this name may be well intended. I haven't looked much into this specific case. Just raising awareness.

Further reading:

https://yarnpkg.com/lang/en/docs/yarn-lock/
https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html

njgheorghita:
@cburgdorf Good to know, thanks. I believe yarn was the original inspiration for using "lockfile" but for the purposes of py-ethpm and the ethpm-spec in general the choice was to rename "lockfile" to "package":
https://github.com/ethpm/ethpm-spec/issues/73

I'm not sure if there's a better way to distinguish between which version of the ethpm-spec the "package" conforms to with the version of the package itself?

ethpm/assets/release-package.schema.v2.json
pipermerriam: Spec Version?

Also we can drop the `release-` prefix on this file since it's not just a "package"

pipermerriam:
@cburgdorf the lockfile name does originally come from that concept but the way the spec evolved, the lockfile name no longer makes sense.  I expect for actual package managers to implement their own lockfiles based on what packages they have installed.

ethpm/assets/release-package.schema.v2.json
njgheorghita: Sounds good to me!

njgheorghita:
@pipermerriam Any thoughts on this CircleCI failure? Yesterday I pinned `rlp<1.0` but that caused a failure earlier today (a dependency wanted 0.4.7 when 0.6.0 was installed if I remember correctly) so I changed it to `rlp==0.4.7` and now circle's upset again. What's the best way to resolve this?

PR #26, opened 2018-05-02
Expand schema to support V2

njgheorghita:
### What was wrong?
- Package schema didn't support some V2 adjustments, like limiting contract name length, bytecode objects, and content-addressed uri's
- Minor changes necessary to update dependencies
- Some general cleanup 

#### Cute Animal Picture
image


ethpm/assets/test_greeter.py
```diff
@@ -1,15 +0,0 @@
-def test_greeter(chain):
```
njgheorghita: I just deleted this file, since it didn't appear to be of much importance - but I could be wrong?

ethpm/assets/package.schema.v2.json
```diff
@@ -2,14 +2,14 @@
...
-    "spec_version",
+    "manifest_version",
     "package_name",
     "version"
   ],
   "version": "2",
   "additionalProperties": false,
   "properties": {
-    "spec_version": {
+    "manifest_version": {
```
njgheorghita: I went with "manifest_version" - seemed like a better descriptor and that's what the guys at truffle went with. https://github.com/trufflesuite/ethpm-js/blob/master/templates/epm.json

ethpm/assets/test_greeter.py
pipermerriam: probably leftover from a copy/paste between projects.

ethpm/assets/package.schema.v2.json
```diff
@@ -2,14 +2,14 @@
   "title": "Release Package Specification",
   "type": "object",
   "required": [
-    "spec_version",
+    "manifest_version",
```
pipermerriam: Assuming this file is largely copy/pasted from the `ethpm-spec` repository, maybe we should include that repository as a submodule to this one and just reference the spec through the submodule.  Should be way easier to stay up-to-date with the main spec that way.

njgheorghita: @pipermerriam not sure how long it'll take the schema in `ethpm-spec` to get updated? is it cool to go ahead and merge this (there's also some unrelated cleanup) - and i'll add referencing `ethpm-spec`'s spec via submodule as an issue here to be done once `ethpm-spec` pr's get merged?

pipermerriam: Go for it.

PR #31, opened 2018-05-05
Use non-comment type annotation

cburgdorf:
### What was wrong?

The comment type annotations are only needed for Python 3.5 support which I assume isn't a goal for this project. Comment type annotations have several disadvantages over proper ones.

### How was it fixed?

Used proper type annotation.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/ethpm/__init__.py
+++ b/ethpm/__init__.py
@@ -11,6 +11,6 @@ if sys.version_info.major < 3:
 
 
 BASE_DIR = os.path.dirname(__file__)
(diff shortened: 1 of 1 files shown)


pipermerriam:
Python 3.5 support is going to be required for this project as it'll end up as a dependency to web3.py which will support python 3.5 for the foreseeable future.

cburgdorf:
Oh. That's unfortunate. Good to know. Need to revert this then. Also need
to set up CI to fail on non 3.5 compliant code.

njgheorghita:
@cburgdorf made the change back to comment types in #23

PR #32, opened 2018-05-05
Add some custom type hints

njgheorghita:
### What was wrong?
Added a `ContractName` and `URI` type to help distinguish from regular `str`s when type checking 

#### Cute Animal Picture
image


njgheorghita:
@cburgdorf @pipermerriam Any rules of thumb for when I should think to reach for a custom type instead of a generic?

eth_typing/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .misc import (
```
cburgdorf: I think we should find a different directory name. It's probably inspired by the `eth_typing` in the `py-evm` project but the reason that is called `eth_typing` there is because it is really intended to be moved out of the project to become a package of shared types that are used across various ethereum projects.

In this particular case though, I think we just need a place to put various types for *this* project. I would vote for creating a directory `typing` directly under `eth_pm`

cburgdorf:
As a rule of thumb: A `NewType` makes sense where you want to semantically differentiate something that you can not differentiate based on the raw shape of it. The canonical example is an API that wants to consume a distance in *miles* which, without a `NewType` is usually represented as an `int`. Calling this API with a distance in *kilometers* will obviously yield faulty results but the type system can't help you as it has no idea what kind of `int` the method should accept. 

I find `URI` a good example (in fact `URI` exists as a proper type in many languages anyway). I think `ContractName` is a bit more controversial but I think we can just try to be restrictive and see how it feels.

eth_typing/__init__.py
njgheorghita: @cburgdorf Yup, was inspired by `py-evm`, but your suggestion makes sense to me.

PR #33, opened 2018-05-10
Minor tweaks to json schema

njgheorghita:
### What was wrong?
- The json schema and packages needed some tweaking to catch up to V2

### How was it fixed?
- Tweaked things

#### Cute Animal Picture
image


ethpm/__init__.py
```diff
@@ -11,6 +11,6 @@
 
 
 BASE_DIR = os.path.dirname(__file__)
-ASSETS_DIR: str = os.path.join(BASE_DIR, 'assets')
+ASSETS_DIR = os.path.join(BASE_DIR, 'assets')  # type: str
```
njgheorghita: change for python 3.5 compliance

ethpm/assets/package.schema.v2.json
```diff
@@ -182,10 +188,11 @@
 	  "title": "Contract Bytecode",
 	  "type": "object",
 	  "required": ["bytecode"],
+	  "additionalProperties": false,
 	  "properties": {
 		"bytecode": {
 		  "type": "string",
-		  "pattern": "^0x[0-9a-fA-Ffff]*$"
+		  "pattern": "^0x[0-9a-zA-Z_]*$"
```
njgheorghita: Made this change since bytecode will have contract names in them? which would require all the chars. Though this might change as `ethpm-spec` #70 gets finalized

ethpm/assets/package.schema.v2.json
```diff
@@ -81,6 +81,7 @@
     "PackageMeta": {
       "title": "Metadata about the package",
       "type": "object",
+	  "additionalProperties": false,
```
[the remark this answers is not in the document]
pipermerriam: Take a look here:

https://github.com/ethpm/ethpm-spec/blob/master/package.spec.md#document-specification

We're currently explicit about allowing extra fields at least at the top level of the package.  I can see an argument for being **very** strict, however, I think that allowing additional fields is a good thing as it allows the spec to be *extended*.  For example, an organization might add an `x-signature` field to the package meta to ship a digital signature around within each of the packages they produce.  Other implementations can easily just ignore such extra fields.

Are there any specific problems you see arising from extra field inclusion?

ethpm/assets/package.schema.v2.json
pipermerriam: I would just make it conform to https://github.com/ethpm/ethpm-spec/pull/70 since the old format is going away and fundamentally broken for certain cases.

ethpm/assets/package.schema.v2.json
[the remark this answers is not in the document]
njgheorghita: Hmmm. . . ok yeah I see that allowing extensions to the spec is a good thing. 

I just got a little spooked when I realized we were validating 4 packages that didn't conform to the spec (fixed on line 52 of the schema). Restricting additional fields would have pointed this out. So, I don't have a specific problem in mind, but w/o the restriction I feel like sloppy mistakes can slip through with all this very easy to read json/regex. 

If an org wanted to add `x-signature` for example, could they simply add the property to the schema definitions and adjust the schema as needed? Or would encouraging schema modifications as needed defeat the purpose of having a standardized schema in the first place?

ethpm/assets/package.schema.v2.json
```diff
@@ -278,14 +288,15 @@
         "settings": {
           "title": "The settings used for compilation",
           "anyOf": [
-            {"$ref": "solc_Settings"}
+            {"$ref": "#/definitions/solc_Settings"}
           ]
         }
       }
     },
     "solc_Settings": {
```
njgheorghita: this snake/camel case also makes me uncomfortable - is there a reason it's like this?

ethpm/assets/package.schema.v2.json
```diff
@@ -163,6 +166,9 @@
         "block": {
           "$ref": "#/definitions/BlockHash"
         },
+		"bytecode": {
```
njgheorghita: Is it safe to assume that if `bytecode` is present, then there will always be a `runtime_bytecode` object present (and vice versa). I know with jsonschema we can add dependencies that can enforce that

ethpm/assets/package.schema.v2.json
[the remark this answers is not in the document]
pipermerriam: The problem with that is that then any tooling (like this one) wouldn't be usable with those packages.  If I'm an organization that's building a private packaging system to be used on an internal blockchain, I'd still really like to be able to use the open source tools.

ethpm/assets/package.schema.v2.json
pipermerriam: not that I recall.  I've been thinking of switching us over to `rAndOMCasInG` though.

ethpm/assets/package.schema.v2.json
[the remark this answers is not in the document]
njgheorghita: Word, alright that makes sense. I'm guessing it's not probable to anticipate all the possible extensions all the developers/orgs might want to implement - so i'll get rid of the restrictions.

ethpm/assets/package.schema.v2.json
njgheorghita: Not horribly opposed to the `raN_DomcAS_ing` - but might cause some heads to explode - and that's no fun for anyone

PR #35, opened 2018-05-14
Validate runtime bytecode utils

njgheorghita:
### What was wrong?
Lack of any functions to handle validation that deployed bytecode of a contract matches the `runtime_bytecode` specified in the package

### How was it fixed?
Wrote util functions (heavily borrowed from populus) to compare and validate bytecodes

#### Cute Animal Picture
image


tests/conftest.py
```diff
@@ -51,6 +56,21 @@ def all_packages():
...
+def package_with_deployed_instance_of_contract(valid_package, w3):
+    with open("ethpm/assets/v2-packages/safe-math-lib/SafeMathLib.sol") as file:
+        contract_source_code = file.read()
+    compiled_sol = compile_source(contract_source_code)
+    contract_interface = compiled_sol['<stdin>:SafeMathLib']
+    SafeMathLib = w3.eth.contract(abi=contract_interface['abi'], bytecode=contract_interface['bin'])
+    tx_hash = SafeMathLib.constructor().transact()
+    tx_receipt = w3.eth.getTransactionReceipt(tx_hash)
+    address = tx_receipt['contractAddress']
+    print(address)
```
njgheorghita: @carver I'd expect every time I run this that I'd get a different address for where the contract was deployed. But i'm getting the exact address every time. Am I missing something? Or is that the expected behavior?

ethpm/utils/bytecode_validation.py
```diff
@@ -0,0 +1,85 @@
...
+    Verify that expected_bytecode matches the bytecode found at address
+    """
+    # Check that the contract has bytecode
+    if expected_bytecode in EMPTY_BYTECODE_VALUES:
+        raise ValueError(
+            "Contract instances which contain an address cannot have empty "
+            "runtime bytecode"
+        )
+
+    chain_bytecode = web3.toHex(web3.eth.getCode(address))
```
njgheorghita: @carver Also here, the code returned from `web3.eth.getCode(address)` always begins with `0x73<address>...`, is that expected? I'm trying to just retrieve the code stored at that address

tests/conftest.py
carver: > every time I run this

Do you mean the same address across test runs? Or the same address in different tests during the same `pytest` run?

Assuming the latter, `w3` (and more importantly, its eth-tester) is reinitialized every time. Since the contract address is based on sending address's nonce, and the nonce is being reset, you will get the same contract address every time.

ethpm/utils/bytecode_validation.py
carver: `getCode` will give you the runtime bytecode, I believe, which does appear to start with a `'0x73'`. Is that a match?

tests/conftest.py
njgheorghita: i'm getting the same address across different `pytest` runs . . .

pipermerriam: @njgheorghita that is expected.  You can check and verify that the `sender/from` and `nonce` for the transaction are the same.  You should only expect different results if you are using a randomized `from` account.  contract address generation logic can be found here: https://github.com/ethereum/py-evm/blob/master/evm/utils/address.py#L12

njgheorghita: @pipermerriam ahh great, good to know thanks. any harm in simply choosing a random account from the 10 automatically generated accounts? so i can skip seeding it with ether for deployment

njgheorghita:
@pipermerriam @carver how would i go about getting solc installed on circle's build?

pipermerriam:
You should be able to find how `py-solc` is used to install solidity if you look in the populus travis.yml file.

You will almost definitely need to update `py-solc` to support installation of a more modern version of solidity.

.circleci/config.yml
```diff
@@ -26,6 +28,18 @@ common: &common
...
+        command: |
+          if [[ -n "$SOLC_VERSION" ]]; then 
+            echo "$SOLC_VERSION"
+            sudo python -m pip install "py-solc>=1.2.0"
+            sudo python -m solc.install $SOLC_VERSION
+            echo 'export SOLC_BINARY=/root/.py-solc/solc-$SOLC_VERSION/bin/solc' >> $BASH_ENV
+            source $BASH_ENV
+            sudo chmod u+x $SOLC_BINARY
+            sudo $SOLC_BINARY --version
+          fi
```
njgheorghita: @pipermerriam @davesque Does anything jump out to you as wrong here? It seems to me like `solc` installs fine - `$SOLC_BINARY` is set correctly for `py-solc` to use - but the test errors out saying `Permission denied` when I try to run `solc.compile`

carver: `sudo chmod u+x $SOLC_BINARY` is only giving execution rights to the root user.

njgheorghita: @carver I also tried `a+rwx` and `-R 755` but neither of those seem to grant the right permissions.

.circleci/config.yml
```diff
@@ -26,9 +28,29 @@ common: &common
...
     - run:
         name: run tox
-        command: ~/.local/bin/tox
+        command: |
+          if [[ -n "$SOLC_VERSION" ]]; then
+            echo "$SOLC_BINARY"
+            sudo $SOLC_BINARY --version
+            $SOLC_BINARY --version
```
njgheorghita: Shouldn't the permissions set on 41 - `all can read, write, execute`, allow 51 to run? 50 is running fine. I've also tried `chmod -R 755 ...` on 41 which didn't help

.circleci/config.yml
```diff
@@ -5,6 +5,8 @@ version: 2.0
 
 common: &common
   working_directory: ~/repo
+  environment:
+    SOLC_VERSION: v0.4.23
```
njgheorghita: hardcoded for now - but will make dynamic down the road to support multiple versions

PR #38, opened 2018-06-01
Write initial base for web3.pm module

njgheorghita:
### What was wrong?
Web3 plans to use py-ethpm under the hood to make it simple to fetch and use packages. This is the initial groundwork to start building the `web3.pm` module. 

### How was it fixed?
Write a `pm` module, attach to a web3 object, and fetch a package. Only file-paths are currently supported. 

#### Cute Animal Picture
image


ethpm/utils/contract.py
```diff
@@ -4,19 +4,25 @@
...
 )
 
+from solc import compile_files
+
+from web3 import Web3
+
+from ethpm.exceptions import ValidationError
+
+
+PACKAGES_DIR = 'ethpm/assets/v2-packages/'
```
pipermerriam: This will break in many cases.  My preferred way to do this is:

```python
import ethpm

BASE_DIR = os.path.dirname(ethpm.__file__)  # gets the root directory that the `ethpm` module is in.

V2_PACKAGES = os.path.joins(BASE_DIR, 'ethpm', 'assets', 'v2-packages')
```

ethpm/utils/contract.py
```diff
@@ -58,3 +64,14 @@ def generate_contract_factory_kwargs(
...
         runtime_bytecode = to_bytes(text=contract_data["runtime_bytecode"]["bytecode"])
         yield "bytecode_runtime", encode_hex(runtime_bytecode)
+
+
+def compile_contracts(contract_name: str, alias: str, paths: List[str]) -> str:
+    '''
+    Compile multiple contracts to bytecode.
+    '''
+    bin_id = '{0}.sol:{0}'.format(contract_name)
+    contract_paths = [PACKAGES_DIR + alias + path[1:] for path in paths]
```
pipermerriam: Instead of `+` which is a little opaque when building strings, I think using string formatting will be easier to read.

```python
[
    "{dir}{alias}{path}".format(dir=PACKAGES_DIR, alias=alias, path=path[1:])
    for path in paths
]
```

setup.py
```diff
@@ -48,14 +48,15 @@
     url='https://github.com/ethpm/py-ethpm',
     include_package_data=True,
     install_requires=[
-        'cytoolz==0.9.0',
+        "cytoolz>=0.9.0,<1.0.0;implementation_name=='cpython'",
```
pipermerriam: to do this you also need to include the `toolz` package for `pypy` installations.

setup.py
```diff
@@ -48,14 +48,15 @@
...
         'eth-keys==0.2.0b3',
         'eth-tester==0.1.0b24',
         'eth-utils==1.0.2',
         'jsonschema==2.6.0',
-        'py-evm==0.2.0a11',
-        'rlp==0.6.0',
-        'web3==4.2.0',
+        'py-evm==0.2.0a16',
+        'py-solc==2.1.0',
```
pipermerriam: pin within a major version range (might as well go ahead and do all of them that way)

ethpm/utils/contract.py
carver: Or the fancy [pathlib](https://docs.python.org/3/library/pathlib.html), like:

```py
from pathlib import Path

BASE_DIR = Path(ethpm.__file__).parent

V2_PACKAGES = BASE_DIR / 'ethpm' / 'assets' / 'v2-packages'
```

setup.py
```diff
@@ -48,14 +48,16 @@
...
+        "toolz>=0.9.0,<1.0.0;implementation_name=='pypy'",
+        "cytoolz>=0.9.0,<1.0.0;implementation_name=='cpython'",
+        'eth-keys>=0.2.0b3',
+        'eth-tester<0.1.0b25,>=0.1.0b24',
+        'eth-utils>=1.0.2',
+        'jsonschema>=2.6.0',
+        'py-evm==0.2.0a16',
+        'py-solc>=2.1.0',
+        'rlp<2.0.0,>=1.0.1',
+        'web3>=4.2.1',
```
carver: These should all be limited within major version ranges, like:

```py
        'web3>=4.2.1,<5',
```

The APIs might change at each major version change.

PR #39, opened 2018-06-01
Pin dependencies to major version release range

njgheorghita:
### What was wrong?
Pin dependencies to major version release range

#### Cute Animal Picture
image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,16 +15,16 @@ readme = open(os.path.join(DIR, 'README.md')).read()
 
 extras_require={
     'test': [
-        'pytest>=3.2.1',
-        'tox>=1.8.0',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -15,16 +15,16 @@
...
     'lint': [
-        'flake8==3.5.0',
+        'flake8>=3.5.0,<4',
         'mypy<0.600',
     ],
     'doc': [
-        'Sphinx>=1.5.5,<2.0.0',
-        'sphinx_rtd_theme>=0.1.9',
+        'Sphinx>=1.5.5,<2',
+        'sphinx_rtd_theme>=0.1.9,<2',
```
carver: This should be `<1` instead.

PR #41, opened 2018-06-11
Bumpversion

njgheorghita:
### What was wrong?
Bumpversion utility for versioning needed 

### How was it fixed?
Added bumpversion with heavy influence from web3.py's implementation

#### Cute Animal Picture
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,12 @@ linux-docs: build-docs
 	xdg-open docs/_build/html/index.html
 
 release: clean
+	CURRENT_SIGN_SETTING=$(git config commit.gpgSign)
+	git config commit.gpgSign true
+	bumpversion $(bump)
+	git push upstream && git push upstream --tags
 	python setup.py sdist bdist_wheel upload
+	git config commit.gpgSign "$(CURRENT_SIGN_SETTING)"
 
 dist: clean
 	python setup.py sdist bdist_wheel
--- a/Makefile
+++ b/Makefile
@@ -49,8 +49,7 @@ release: clean
 	git config commit.gpgSign true
 	bumpversion $(bump)
 	git push upstream && git push upstream --tags
-	python setup.py sdist bdist_wheel
-	twine upload dist/*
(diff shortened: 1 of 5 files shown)


njgheorghita:
I had some tough auth problems which resulted in many unnecessary bumps. Is it ok to leave like this? Or should they be rolled back?

Makefile
```diff
@@ -45,7 +45,13 @@ linux-docs: build-docs
 	xdg-open docs/_build/html/index.html
 
 release: clean
-	python setup.py sdist bdist_wheel upload
+	CURRENT_SIGN_SETTING=$(git config commit.gpgSign)
+	git config commit.gpgSign true
+	bumpversion $(bump)
+	git push upstream && git push upstream --tags
+	python setup.py sdist bdist_wheel
+	twine upload dist/*
```
njgheorghita: I also used `Twine` to upload the dists to pypi b/c of this auth problem - which is about the only diversion from web3's implementation. I tried just about everything here to fix it: https://github.com/pypa/setuptools/issues/941. It works fine now, but I''m not sure if this might be problematic down the road?

davesque: I recently modified the project template to use twice because I was having similar issues.  So using twine should be fine I think.

pipermerriam: No objections, can you however, explain what the problems you were running into were?

njgheorghita: Without Twine - when it came time to upload the dist to pypi - it would ask me for my pypi password (without having asked for my username)- I tried many times with the correct password always got an unable to authenticate error. I also tried creating a `.pypirc` with my username and password in it - a la the link ^^^ - but that didn't cause any different behavior when running the command

pipermerriam: Try putting this in `$HOME/.pypirc`

```
[distutils]
index-servers =
    pypi

[pypi]
username:pipermerriam
```

njgheorghita: Nice! It all works now without twine - I put `.pypirc` in `py-ethpm`'s root which I'm guessing is the reason it wasn't picking up the config info

PR #43, opened 2018-06-14
Reference example packages via submodule to ethpm-spec

njgheorghita:
### What was wrong?
Rather than duplicating package management examples in this repo, reference the standard examples located in `ethpm-spec` via a submodule

### How was it fixed?
- Removed `assets/` directory
- Included `ethpm-spec` as a submodule
- Adjusted paths where necessary

#### Cute Animal Picture
image


ethpm/__init__.py
```diff
@@ -10,8 +11,9 @@
     warnings.warn(warn_msg, DeprecationWarning)
 
 
-BASE_DIR = os.path.dirname(__file__)
-ASSETS_DIR = os.path.join(BASE_DIR, 'assets')  # type: str
-V2_PACKAGES = os.path.join(ASSETS_DIR, 'v2-packages')
+BASE_DIR = Path('.')
```
pipermerriam: Instead of `'.'` you can use `__file__` to get the path to *this* file, and then navigate from there.  I believe that `'.'` won't reliably map to the root dir here.

PR #44, opened 2018-06-19
Move filesystem util to utils

njgheorghita:
### What was wrong?
Minor reorganization of util to it's own file in `utils` module

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/utils/filesystem.py
+++ b/ethpm/utils/filesystem.py
@@ -1,0 +1,23 @@
+import json
+
+from typing import (
+    Dict,
+    IO,
+)
+
+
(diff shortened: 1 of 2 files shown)


ethpm/utils/filesystem.py
```diff
@@ -0,0 +1,23 @@
+import json
+
+from typing import (
+    Dict,
+    IO,
+)
+
+
+def _load_package_data_from_file(file_obj: IO[str]) -> Dict[str, str]:
```
pipermerriam: You can drop the `_` prefix since this now lives in the `utils` namespace.

PR #45, opened 2018-06-20
Write registry uri validation utils

njgheorghita:
### What was wrong?
Fetching a manifest content-addressed URI from a package requires a registry URI scheme. These are the utils to validate whether or not a URI has been properly formatted. 

### How was it fixed?
A lot of the rules imposed here are followed by convention set by the dependent libraries. 
i.e.
- enforcing "://" to follow the `scheme` is convention set in `urllib`

* Using `ercxxx` as the scheme will be updated once the registry contract standard makes it's way through the ERC process

#### Cute Animal Picture
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,8 +126,8 @@ scheme://authority/package-name/version
 
 * URI must be a string type
 * `scheme`: `ercxxx` 
-* `authority`: Must be a valid ENS domain or a valid hex address pointing towards a registry contract.
+* `authority`: Must be a valid ENS domain or a valid checksum address pointing towards a registry contract.
 * `package-name`: Must conform to the package-name as specified in the [EthPM-Spec](http://ethpm-spec.readthedocs.io/en/latest/package-spec.html#package-name).
-* `version`: Must be a valid semver string as specified [here](https://semver.org/).
+* `version`: The URI escaped version string.
 
 i.e. `ercxxx://packages.zeppelinos.eth/owned/1.0.0`
--- a/README.md
+++ b/README.md
@@ -114,3 +114,20 @@ The `Package` class should provide access to the full dependency tree.
 ```sh
 git submodule init
 ```
+
+
+## Registry URI 
+
+The URI to lookup a package from a registry should follow the following format. (subject to change as the Registry Contract Standard makes it's way through the EIP process)
+
+```
+scheme://authority/package-name/version
+```
+
+* URI must be a string type
+* `scheme`: `ercxxx` 
+* `authority`: Must be a valid ENS domain or a valid hex address pointing towards a registry contract.
+* `package-name`: Must conform to the package-name as specified in the [EthPM-Spec](http://ethpm-spec.readthedocs.io/en/latest/package-spec.html#package-name).
+* `version`: Must be a valid semver string as specified [here](https://semver.org/).
+
+i.e. `ercxxx://packages.zeppelinos.eth/owned/1.0.0`
--- a/README.md
+++ b/README.md
@@ -121,13 +121,13 @@ git submodule init
 The URI to lookup a package from a registry should follow the following format. (subject to change as the Registry Contract Standard makes it's way through the EIP process)
 
 ```
-scheme://authority/package-name/version
+scheme://authority/package-name?version=x.x.x
 ```
 
 * URI must be a string type
 * `scheme`: `ercxxx` 
 * `authority`: Must be a valid ENS domain or a valid checksum address pointing towards a registry contract.
 * `package-name`: Must conform to the package-name as specified in the [EthPM-Spec](http://ethpm-spec.readthedocs.io/en/latest/package-spec.html#package-name).
-* `version`: The URI escaped version string.
+* `version`: The URI escaped version string, *should* conform to the [semver](http://semver.org/) version numbering specification.
 
-i.e. `ercxxx://packages.zeppelinos.eth/owned/1.0.0`
+i.e. `ercxxx://packages.zeppelinos.eth/owned?version=1.0.0`
--- a/ethpm/exceptions.py
+++ b/ethpm/exceptions.py
@@ -5,6 +5,12 @@ class InsufficientAssetsError(Exception):
     """
 
 
+class UriNotSupportedError(Exception):
+    """
+    Error to signal a URI scheme is not supported.
+    """
+
+
 class ValidationError(Exception):
     """
     Error to signal something does not pass a validation check.
--- a/ethpm/exceptions.py
+++ b/ethpm/exceptions.py
@@ -1,18 +1,20 @@
 class InsufficientAssetsError(Exception):
     """
-    Error to signal a Manifest or Package does not
+    Raised when a Manifest or Package does not
     contain the required assets to do something.
     """
-
-
-class UriNotSupportedError(Exception):
-    """
-    Error to signal a URI scheme is not supported.
-    """
+    pass
 
 
 class ValidationError(Exception):
     """
-    Error to signal something does not pass a validation check.
+    Raised when something does not pass a validation check.
+    """
+    pass
+
+
+class UriNotSupportedError(ValidationError):
+    """
+    Raised when URI scheme does not conform to the registry URI scheme.
     """
     pass
--- a/ethpm/utils/registry.py
+++ b/ethpm/utils/registry.py
@@ -1,0 +1,87 @@
+import re
+import semver
+
+from urllib import parse
+
+from typing import List
+
+from eth_utils import (
+    is_hex_address,
+)
+
+from ethpm.exceptions import UriNotSupportedError
+
+
+# TODO: update with correct scheme once ERC is completed
+REGISTRY_URI_SCHEME = 'ercxxx'
+PACKAGE_NAME_REGEX = '[a-zA-Z][-_a-zA-Z0-9]{0,255}'
+
+
+def is_valid_authority(auth: str) -> None:
+    """
+    Raises exception if authority is not a valid ENS domain or contract address.
+    """
+    if is_ens_domain(auth) is False and not is_hex_address(auth):
+        raise UriNotSupportedError('{0} is not a valid registry URI authority.'.format(auth))
+
+
+def is_ens_domain(auth: str) -> bool:
+    """
+    Returns false if URI authority is not a valid ENS domain.
+    """
+    if auth[-4:] != '.eth' or len(auth.split('.')) not in [2, 3]:
+        return False
+    return True
+
+
+def is_valid_semver(pkg_version: str) -> None:
+    """
+    Raises exception if package version is not valid semver.
+    """
+    try:
+        semver.parse(pkg_version)
+    except ValueError as exc:
+        raise UriNotSupportedError('{0} is not a valid registry URI version.'.format(pkg_version))
+
+
+def is_valid_package_name(pkg_name: str) -> None:
+    """
+    Returns boolean if the value is a valid package name.
+    """
+    if not bool(re.match(PACKAGE_NAME_REGEX, pkg_name)):
+        raise UriNotSupportedError('{0} is not a valid package name.'.format(pkg_name))
+
+
+def is_valid_scheme(scheme: str) -> None:
+    """
+    Raises exception is package scheme is not a valid scheme.
+    """
+    if scheme != REGISTRY_URI_SCHEME:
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -114,3 +114,20 @@ The `Package` class should provide access to the full dependency tree.
...
+
+```
+scheme://authority/package-name/version
+```
+
+* URI must be a string type
+* `scheme`: `ercxxx` 
+* `authority`: Must be a valid ENS domain or a valid hex address pointing towards a registry contract.
+* `package-name`: Must conform to the package-name as specified in the [EthPM-Spec](http://ethpm-spec.readthedocs.io/en/latest/package-spec.html#package-name).
+* `version`: Must be a valid semver string as specified [here](https://semver.org/).
```
pipermerriam: I don't think we can require semver.  I think this should just be something like:

> The URI escaped version string.

README.md
```diff
@@ -114,3 +114,20 @@ The `Package` class should provide access to the full dependency tree.
...
+
+The URI to lookup a package from a registry should follow the following format. (subject to change as the Registry Contract Standard makes it's way through the EIP process)
+
+```
+scheme://authority/package-name/version
+```
+
+* URI must be a string type
+* `scheme`: `ercxxx` 
+* `authority`: Must be a valid ENS domain or a valid hex address pointing towards a registry contract.
```
pipermerriam: Requiring checksummed addresses is probably an ok thing for us to do.

ethpm/utils/registry.py
```diff
@@ -0,0 +1,87 @@
...
+    except ValueError as exc:
+        raise UriNotSupportedError('{0} is not a valid registry URI version.'.format(pkg_version))
+
+
+def is_valid_package_name(pkg_name: str) -> None:
+    """
+    Returns boolean if the value is a valid package name.
+    """
+    if not bool(re.match(PACKAGE_NAME_REGEX, pkg_name)):
+        raise UriNotSupportedError('{0} is not a valid package name.'.format(pkg_name))
```
pipermerriam: The naming of the function would make me surprised when this raised an exception as I would expect it to return a boolean.  Maybe move these *validation* functions to a separate module.  I've found it good to have two utils for each of these things.

- `is_valid_thing(value) -> bool` (for *checking*)
- `validate_thing(value) -> None` (either raises or returns `None` for validation)

README.md
njgheorghita: Why not? Just to leave it open-ended in case someone wants to use a different versioning scheme? Should I be doing any kind of validation on the version?

README.md
```diff
@@ -114,3 +114,20 @@ The `Package` class should provide access to the full dependency tree.
...
+
+```
+scheme://authority/package-name/version
+```
+
+* URI must be a string type
+* `scheme`: `ercxxx` 
+* `authority`: Must be a valid ENS domain or a valid checksum address pointing towards a registry contract.
+* `package-name`: Must conform to the package-name as specified in the [EthPM-Spec](http://ethpm-spec.readthedocs.io/en/latest/package-spec.html#package-name).
+* `version`: The URI escaped version string.
```
pipermerriam: Ok, responding to your other comment here about *why*.

1. We want to be flexible to allow alternate version schemes, while probably only supporting semver for anything other than exact version matching.
2. This means this spec needs to be expanded a bit to allow specifying things like `>=1.0,<2` (which I believe is *either* valid semver, or a valid semver abstraction.

So for example (without proper url escaping)

- `ercXXX://packages.augur.eth/safe-math/>=1.0.2,<2`

When I say expanded, I mean we should define the term "version string" loosely and put some sort of recommendation that tooling support this style of semver.  I'd recommend looking for an RFC somewhere about this, and just pointing to that.

README.md
```diff
@@ -114,3 +114,20 @@ The `Package` class should provide access to the full dependency tree.
...
 git submodule init
 ```
+
+
+## Registry URI 
+
+The URI to lookup a package from a registry should follow the following format. (subject to change as the Registry Contract Standard makes it's way through the EIP process)
+
+```
+scheme://authority/package-name/version
```
pipermerriam: thoughts on some alternate formats which I don't necessarily think are better but worth exploring other options.

- `scheme://package-name@authority/version`
- `scheme://package-name@authority?version=version` (less easy to type out by hand because of query parameter escaping)
- `scheme://authority/package-name?version=version` (less easy to type out by hand because of query parameter escaping)

In all three of these examples, the distinction between `package-name` and `version` is clearer as opposed to having to do string manipulation to separate them (as they are in distinctly different parts of the url scheme).

> I think in all of these, version should be optional, if left off, the *latest* version would be used.

ethpm/utils/registry.py
```diff
@@ -0,0 +1,48 @@
...
+    validate_registry_uri_scheme,
+    validate_registry_uri_version,
+)
+
+
+def is_ens_domain(auth: str) -> bool:
+    """
+    Returns false if URI authority is not a valid ENS domain.
+    """
+    if auth[-4:] != '.eth' or len(auth.split('.')) not in [2, 3]:
```
pipermerriam: This is sufficiently cryptic to read that it deserves a comment.  I am still not sure what the second check is validating.

ethpm/utils/registry.py
```diff
@@ -0,0 +1,48 @@
...
+
+from ethpm.validation import (
+    validate_package_name,
+    validate_registry_uri_authority,
+    validate_registry_uri_scheme,
+    validate_registry_uri_version,
+)
+
+
+def is_ens_domain(auth: str) -> bool:
```
pipermerriam: maybe go ahead and use the full name `authority`, otherwise the assumption is likely to be this stands for `authorization`

ethpm/utils/registry.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def parse_path(path: str) -> List[str]:
+    """
+    Raises exception if path is empty string.
+    If path is valid, return the name and version
+    """
+    if len(path.split('/')) not in [2, 3, 4]:
+        raise UriNotSupportedError('{0} is not a valid registry URI path.'.format(path))
+    return path.split('/')[1:3]
```
pipermerriam: The `version` part should be unescaped before returning it as it could contain escaped characters (though I think it's better for us to find an existing RFC for version strings and use that to inform our decisions here.

ethpm/exceptions.py
```diff
@@ -5,6 +5,12 @@ class InsufficientAssetsError(Exception):
     """
 
 
+class UriNotSupportedError(Exception):
```
pipermerriam: The way this is used, it looks like it should inherit from `ValidationError`

ethpm/utils/registry.py
```diff
@@ -0,0 +1,48 @@
...
+def is_ens_domain(auth: str) -> bool:
+    """
+    Returns false if URI authority is not a valid ENS domain.
+    """
+    if auth[-4:] != '.eth' or len(auth.split('.')) not in [2, 3]:
+        return False
+    return True
+
+
+def is_registry_uri(uri: str) -> bool:
```
pipermerriam: If you split this into two functions then you'll be able to resolve the circular import issue your having.  This function either returns `True` or throws an exception which violates the principle of least surprise.  This should instead use all of the `is_` checks to return `True` or `False` and then write a `validate_registry_uri` function in your `validation.py` utils file that uses this.

Should allow you to remove all of the imports from this module that import from `validation` and to just have `validation` import from this module.


https://en.wikipedia.org/wiki/Principle_of_least_astonishment

davesque:
> `scheme://package-name@authority/version`
> `scheme://package-name@authority?version=version` (less easy to type out by hand because of query parameter escaping)
> `scheme://authority/package-name?version=version` (less easy to type out by hand because of query parameter escaping)

The only alternative option I really want to get behind is the third one, which has the version as a query param.  The others seem to violate the generic syntax conventions of URIs and place non-auth data in the "userinfo" component (https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Generic_syntax).

The query param option seems maybe the most correct since a version number really does act like a filter.  Although that sort of implies that you'd specify version ranges with something like this:
```
scheme://authority/<package-name>?version_gte=1.0.2&version_lt=2
```
Otherwise, you'd end up with something pretty ugly when you try to encode a version range in a single param:
```
scheme://authority/<package-name>?version=%3E%3D1.0.2%2C%3C2

">=1.0.2,<2" is "%3E%3D1.0.2%2C%3C2" when URI encoded
```
Would it make sense to say that the URI just doesn't include any version information?  Then, it would be up to the package library to figure out what version you want:
```
pkg = w3.pm.get_package_from_uri('ens://packages.ethereum.eth/greeter', '1.0.0')
pkg = w3.pm.get_package_from_uri('ens://packages.ethereum.eth/greeter', '>=1.0.2,<2')
```

njgheorghita:
I'm a fan of enforcing versions to be submitted as params (not crazy about the param names (i.e. `version_gte`,`version_lt`, ...) but I can't think of anything better. 

> scheme://authority/<package-name>?version_gte=1.0.2&version_lt=2

> Would it make sense to say that the URI just doesn't include any version information?

I'm not sure about this - I feel like i might want to be able to copy&paste a URI and send it to somebody and be sure it's resolving to the right package - so i'd like to see all the data contained in the URI.

pipermerriam:
We can support both.

```python
# we pick one of these as shorthand
w3.pm.get_package_from_uri('ens://packages.ethereum.eth/greeter', params={'version': '>=1.0.2,<2'})
w3.pm.get_package_from_uri('ens://packages.ethereum.eth/greeter', version='>=1.0.2,<2')
w3.pm.get_package_from_uri('ens://packages.ethereum.eth/greeter', '>=1.0.2,<2')

# which ends up equivalent to
w3.pm.get_package_from_uri('ens://packages.ethereum.eth/greeter?version=%3E%3D1.0.2%2C%3C2`)
```

That should be very easy to both support and validate.

ethpm/exceptions.py
```diff
@@ -3,16 +3,18 @@ class InsufficientAssetsError(Exception):
...
     contain the required assets to do something.
     """
+    pass
 
 
-class UriNotSupportedError(Exception):
+class ValidationError(Exception):
     """
-    Error to signal a URI scheme is not supported.
+    Error to signal something does not pass a validation check.
```
cburgdorf: I would suggest to adopt the `Raised when` scheme for exception doc strings that we adopted from the Python standard lib for Py-EVM/Trinity.

image

ethpm/exceptions.py
```diff
@@ -3,16 +3,18 @@ class InsufficientAssetsError(Exception):
...
     """
+    pass
 
 
-class ValidationError(Exception):
+class UriNotSupportedError(ValidationError):
     """
-    Error to signal something does not pass a validation check.
+    Error to signal a URI scheme is not supported.
```
cburgdorf: See above

ethpm/exceptions.py
njgheorghita: Dope, thanks for the tip!

ethpm/utils/registry.py
```diff
@@ -1,43 +1,20 @@
...
-from ethpm.validation import (
-    validate_package_name,
-    validate_registry_uri_authority,
-    validate_registry_uri_scheme,
-    validate_registry_uri_version,
-)
-
 
-def is_ens_domain(auth: str) -> bool:
+def is_ens_domain(authority: str) -> bool:
```
cburgdorf: Not part of this PR but in general the docstrings should use imperative, present tens.

Helper: When the `is_ens_domain` is called it will:

>Return `False` if URI authority is not a valid ENS domain.

ethpm/validation.py
```diff
@@ -18,11 +25,24 @@ def validate_package_name(pkg_name: str) -> None:
         raise UriNotSupportedError('{0} is not a valid package name.'.format(pkg_name))
 
 
+def validate_registry_uri(uri: str) -> None:
+    """
+    Raises exception if URI does not conform the the registry URI scheme.
```
cburgdorf: >Validate a registry uri. Raise exception if URI does not conform the registry URI scheme

njgheorghita: IDK - starting with "Validate a registry URI" seems slightly redundant to me - since it's almost word for word the name of the function. For you, does it help with clarity? I'm thinking ...
When the `validate_registry_uri` is called it will:
> Raise an exception if the URI does not conform to the registry URI scheme.

cburgdorf: Sounds good :+1:

PR #46, opened 2018-06-22
Fetch URI from registry

njgheorghita:
### What was wrong?
`Package` object should be able to be instantiated via a Registry URI scheme. 

### How was it fixed?
Add a class method `Package.from_registry(registry_uri, w3)`. 
Follows the Registry URI scheme as defined in #45 

#### Cute Animal Picture
image


ethpm/exceptions.py
```diff
@@ -1,12 +1,20 @@
 class InsufficientAssetsError(Exception):
```
cburgdorf: @njgheorghita @davesque not for this PR but I think `ethpm` should follow the same strategy as py-evm/trinity in that all exceptions inherit from some common application base exception (e.g. `BaseEthPMException`)

https://github.com/ethereum/py-evm/blob/82022a78a8b2419fc090af21428f20e2c01172f4/trinity/exceptions.py#L4

ethpm/validation.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    Raise an exception if the value is not a valid package name
+    as defined in the EthPM-Spec.
+    """
+    if not bool(re.match(PACKAGE_NAME_REGEX, pkg_name)):
+        raise UriNotSupportedError('{0} is not a valid package name.'.format(pkg_name))
+
+
+def validate_registry_uri(uri: str) -> None:
+    """
```
cburgdorf: I love that you focus on landing the APIs with docs right from the start :heart:

ethpm/constants.py
```diff
@@ -0,0 +1,99 @@
...
+        index[_name] = Package(_uri, _version);
+    }
+
+    function lookupPackage(bytes32 _name) public view returns (string) {
+        return index[_name].uri;
+    }
+
+}
+'''
+REGISTRY_ABI = '''
```
pipermerriam: Same thought here.  Consider moving to `ethpm/assets/registry_abi.json` and load it from that file here.

ethpm/package.py
```diff
@@ -128,6 +131,18 @@ def from_ipfs(cls, ipfs_uri: str) -> 'Package':
 
         return cls(package_data)
 
+    @classmethod
+    def from_registry(cls, registry_uri: str, w3: Web3) -> 'Package':
```
pipermerriam: Don't we have a URI type in `eth-typing`?

ethpm/utils/registry.py
```diff
@@ -0,0 +1,34 @@
...
+    return True
+
+
+def lookup_manifest_uri_located_at_registry_uri(registry_uri: str, w3: Web3) -> str:
+    """
+    Registry must conform to the ERCXXX.
+    """
+    parsed = parse.urlparse(registry_uri)
+    authority = parsed.netloc
+    pkg_name = to_bytes(text=parsed.path[1:])
```
pipermerriam: why the `[1:]`?

ethpm/utils/uri.py
```diff
@@ -0,0 +1,35 @@
...
+INTERNET_PREFIXES = ['http', 'https']
+
+SWARM_PREFIXES = ['bzz', 'bzz-immutable', 'bzz-raw']
+
+
+def get_manifest_from_content_addressed_uri(uri: str) -> Dict[str, Any]:
+    parse_result = parse.urlparse(uri)
+    scheme = parse_result.scheme
+
+    if scheme == 'ipfs':
```
pipermerriam: Might as well move `ipfs` to a constant `IPFS_SCHEME`.

ethpm/utils/uri.py
```diff
@@ -0,0 +1,35 @@
...
+from urllib import parse
+
+from ethpm.exceptions import UriNotSupportedError
+
+from ethpm.utils.ipfs import (
+    fetch_ipfs_package,
+)
+
+
+INTERNET_PREFIXES = ['http', 'https']
```
pipermerriam: *naming nitpick*

`INTERNET_SCHEMES`

ethpm/validation.py
```diff
@@ -0,0 +1,64 @@
...
+    if scheme != REGISTRY_URI_SCHEME:
+        raise UriNotSupportedError('{0} is not a valid registry URI scheme.'.format(scheme))
+
+
+def validate_registry_uri_version(query: str) -> None:
+    """
+    Raise an exception if the version param is malformed.
+    """
+    if 'version=' != query[:8]:
```
pipermerriam: You should use [`urllib.parse.parse_qs`](https://docs.python.org/3.0/library/urllib.parse.html#urllib.parse.parse_qs) to parse the querystring and interact with it as a dictionary.

tests/ethpm/test_package_init_from_registry.py
```diff
@@ -0,0 +1,64 @@
...
+
+from ethpm.exceptions import UriNotSupportedError
+
+from ethpm.constants import REGISTRY_SOURCE
+
+
+# mock out http req to IPFS gateway
+# `fetch_ipfs_package` returns local 'owned' pkg
+@pytest.fixture(autouse=True)
+def mock_request(monkeypatch, owned_manifest):
```
pipermerriam: Just leaving a note that we should be able to drop this mocking once the IPFS fetching is done with a backend system, as we can just swap in our own dummy backend that just returns the content from memory or something like that.

ethpm/validation.py
njgheorghita: Danke!! learning from the best

ethpm/package.py
njgheorghita: not yet in a released version - but it's there in an approved pr

ethpm/utils/registry.py
pipermerriam: maybe consider something like `parsed.path.strip('/')` so that it will take care of trailing slashes too.

tests/ethpm/test_package_init_from_registry.py
njgheorghita: sounds good - added it to the ipfs issue

davesque:
Very cool work, Nick.  I left a couple of comments but mostly this looks like a great start on this functionality.

njgheorghita:
Thanks @davesque! not sure why, but I can't see the comments you left? 🤷‍♂️

davesque:
@njgheorghita I think it was because I hadn't "submitted" my review.  Confusing. :/

Issue #37, opened 2018-05-22
Add configurable IPFS URI backend 


njgheorghita:
### What was wrong?
Right now we only support fetching packages on IPFS, and only through the INFURA gateway.

Also extend backend so it can swap in a dummy backend that will serve content rather than monkeypatching http requests.  (see piper's comment here . . . https://github.com/ethpm/py-ethpm/pull/46)

### How can it be fixed?
Add configurable option to allow fetching packages from other gateways & local nodes

njgheorghita:
@pipermerriam Can you expand a little more on the API you have in mind for the different configurable backends? Are you thinking something similar to the backends used in Populus package management?

pipermerriam:
```python
from abc import ABC, abstract_method

class BaseIPFSBackend(ABC):
    @abstractmethod
    def fetch_ipfs_uri(self, uri: str) -> str:  # or maybe `bytes`?
        """
        Returns the file contents?
        """
        pass

class IPFSOverHTTPBackend(BaseIPFSBackend):
        def __init__(self, base_uri):
            self.base_uri = base_uri

        def fetch_ipfs_uri(self, uri):
            ipfs_hash = extract_hash_from_uri(uri)
            gateway_uri = self.base_uri + '/' + ipfs_hash
            response = requests.get(gateway_uri)
            return response.content

class LocalIPFSBackend(BaseIPFSBackend):
    ... # knows how to connect to a locally running IPFS node...

class InfuraIPFSBackend(IPFSOverHTTPBackend):
    ... # knows the infura IPFS base uris...

class IPFSGatewayBackend(IPFSOverHTTPBackend):
    ... # knows the URI for the IPFS http gateway
```

Something like that.  Then the same plumbing type code like this from py-evm https://github.com/ethereum/py-evm/blob/master/evm/db/__init__.py for loading the backend class, probably configurable via environment variable.

If we want to be **really** generic with URIs (which we do) we probably need a method like:

```python
class BaseURIBackend:
    # generic backend that all URI backends come from
    def can_handle_uri(self, uri) -> bool:
        ... # returns `True` or `False` for whether this backend class can handle the given URI

# and then some glue code in the core logic that fetches URIs


def get_backends_for_uri(backends, uri):
    return tuple(backend for backend in backends if backend.can_handle_uri(uri))

# and then some code to loop over the *good* backends and return the fetched contents.  Maybe an exeption that the backends can raise if for some reason they find they can't handle it.

for backend in good_backends:
[...]

PR #51, opened 2018-06-27
Pluggable URI & IPFS backend class

njgheorghita:
### What was wrong?
Pluggable backend classes were needed for easier development and testing.

### How was it fixed?
Removed existing `fetch_from_ipfs` scheme and implemented a URI backend class a la discussion in #37 

- Created `BaseURIBackend` and various subclasses according to different URI schemes

#### Cute Animal Picture
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,7 @@ The `Package` class should provide access to the full dependency tree.
 git submodule init
 ```
 
+
 ## Registry URI 
 
 The URI to lookup a package from a registry should follow the following format. (subject to change as the Registry Contract Standard makes it's way through the EIP process)
@@ -134,6 +135,15 @@ scheme://authority/package-name?version=x.x.x
 * `version`: The URI escaped version string, *should* conform to the [semver](http://semver.org/) version numbering specification.
 
 i.e. `ercxxx://packages.zeppelinos.eth/owned?version=1.0.0`
+
+
+## URI Backend Class
+* URI Backend Class defaults to `ethpm.backends.ipfs.IFPSGatewayBackend`
+* URI Backend Class is configurable via the environment variable `ETHPM_URI_BACKEND_CLASS` set to a dotted module path
+* `DummyIPFSBackend` is used to avoid http requests within testing
+	* Can accept both a ...
+		* Valid IPFS URI -> `safe-math-lib` manifest
+		* Path to existing manifest/contract in `V2_PACKAGES_DIR` -> defined manifest/contract 
 
 
 ## Release setup
--- a/ethpm/backends/__init__.py
+++ b/ethpm/backends/__init__.py
@@ -2,15 +2,11 @@ import os
 
 from typing import Type
 
-from ethpm.backends.base import (
-    BaseURIBackend,
-)
-from ethpm.utils.module_loading import (
-    import_string,
-)
+from ethpm.backends.base import BaseURIBackend
+from ethpm.utils.module_loading import import_string
 
 
-DEFAULT_URI_BACKEND = 'ethpm.backends.ipfs.IPFSGatewayBackend'
+DEFAULT_URI_BACKEND = "ethpm.backends.ipfs.IPFSGatewayBackend"
 
 
 def get_uri_backend(import_path: str = None) -> BaseURIBackend:
@@ -23,8 +19,5 @@ def get_uri_backend(import_path: str = None) -> BaseURIBackend:
 
 def get_uri_backend_class(import_path: str = None) -> Type[BaseURIBackend]:
     if import_path is None:
-        import_path = os.environ.get(
-            'ETHPM_URI_BACKEND_CLASS',
-            DEFAULT_URI_BACKEND,
-        )
+        import_path = os.environ.get("ETHPM_URI_BACKEND_CLASS", DEFAULT_URI_BACKEND)
     return import_string(import_path)
--- a/ethpm/backends/__init__.py
+++ b/ethpm/backends/__init__.py
@@ -1,0 +1,30 @@
+import os
+
+from typing import Type
+
+from ethpm.backends.base import (
+    BaseURIBackend,
+)
+from ethpm.utils.module_loading import (
+    import_string,
+)
+
+
+DEFAULT_URI_BACKEND = 'ethpm.backends.ipfs.IPFSGatewayBackend'
+
+
+def get_uri_backend(import_path: str = None) -> BaseURIBackend:
+    """
+    Return the `BaseURIBackend` class specified by import_path, default, or env variable.
+    """
+    backend_class = get_uri_backend_class(import_path)
+    return backend_class()
+
+
+def get_uri_backend_class(import_path: str = None) -> Type[BaseURIBackend]:
+    if import_path is None:
+        import_path = os.environ.get(
+            'ETHPM_URI_BACKEND_CLASS',
+            DEFAULT_URI_BACKEND,
+        )
+    return import_string(import_path)
--- a/ethpm/backends/base.py
+++ b/ethpm/backends/base.py
@@ -1,37 +1,34 @@
-from abc import ABC, abstractmethod
+from abc import (
+    ABC,
+    abstractmethod,
+)
 
-from ethpm.utils.ipfs import extract_ipfs_path_from_uri
 
-
-# generic backend that all URI backends come from
 class BaseURIBackend(ABC):
-    def can_handle_uri(self, uri) -> bool:
-        """
-        Returns bool for whether this backend class can handle the given URI.
-        """
-        pass
+    """
+    Generic backend that all URI backends are subclassed from.
 
-    def fetch_uri_contents(uri):
+    All subclasses must implement:
+    can_handle_uri, fetch_uri_contents, base_uri
+    """
+    @abstractmethod
+    def can_handle_uri(self, uri: str) -> bool:
+        """
+        Return a bool indicating whether this backend class can handle the given URI.
+        """
+        raise NotImplementedError("Must be implemented by subclass.")
+
+    @abstractmethod
+    def fetch_uri_contents(self, uri: str) -> bytes:
         """
         Fetch the contents stored at a URI.
         """
-        pass
+        raise NotImplementedError("Must be implemented by subclass.")
 
-
-# glue code in the core logic that fetches URIS
-def get_backends_for_uri(backends, uri):
-    return tuple(backend for backend in backends if backend.can_handle_uri(uri))
-
-
-# and then some code to loop over the *good* backends and return the fetched contents
-# maybe an exception that the backends can raise if for some reason they find that can't handle it
-def get_uri_contents(): 
-    for backend in good_backends:
-        try:
-            contents = backend.fetch_uri_contents(uri)
-        except CannotHandleURI: # custom exception
-            continue
-        else:
-            return contents
-    else:
-        raise SomethingError("No backends were able to fetch the URI contents")
+    @property
+    @abstractmethod
+    def base_uri(self) -> str:
+        """
+        Return the base URI unique to a particular scheme.
+        """
+        raise NotImplementedError("Must be implemented by subclass.")
--- a/ethpm/backends/base.py
+++ b/ethpm/backends/base.py
@@ -1,7 +1,4 @@
-from abc import (
-    ABC,
-    abstractmethod,
-)
+from abc import ABC, abstractmethod
 
 
 class BaseURIBackend(ABC):
@@ -9,26 +6,19 @@ class BaseURIBackend(ABC):
     Generic backend that all URI backends are subclassed from.
 
     All subclasses must implement:
-    can_handle_uri, fetch_uri_contents, base_uri
+    can_handle_uri, fetch_uri_contents
     """
+
     @abstractmethod
     def can_handle_uri(self, uri: str) -> bool:
         """
         Return a bool indicating whether this backend class can handle the given URI.
         """
-        raise NotImplementedError("Must be implemented by subclass.")
+        pass
 
     @abstractmethod
     def fetch_uri_contents(self, uri: str) -> bytes:
         """
         Fetch the contents stored at a URI.
         """
-        raise NotImplementedError("Must be implemented by subclass.")
-
-    @property
-    @abstractmethod
-    def base_uri(self) -> str:
-        """
-        Return the base URI unique to a particular scheme.
-        """
-        raise NotImplementedError("Must be implemented by subclass.")
+        pass
--- a/ethpm/backends/base.py
+++ b/ethpm/backends/base.py
@@ -1,0 +1,37 @@
+from abc import ABC, abstractmethod
+
+from ethpm.utils.ipfs import extract_ipfs_path_from_uri
+
+
+# generic backend that all URI backends come from
+class BaseURIBackend(ABC):
+    def can_handle_uri(self, uri) -> bool:
+        """
+        Returns bool for whether this backend class can handle the given URI.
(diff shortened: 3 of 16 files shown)


tests/ethpm/backends/test_ipfs_backends.py
```diff
@@ -0,0 +1,50 @@
...
+
+from eth_utils import to_text
+
+from ethpm.backends.ipfs import (
+    DummyIPFSBackend,
+    InfuraIPFSBackend,
+    IPFSGatewayBackend,
+)
+
+
```
njgheorghita: Not quite sure how to test `IPFSGatewayBackend` and `InfuraIPFSBackend` without making HTTP requests.

tests/ethpm/test_package_init_from_registry.py
```diff
@@ -39,20 +29,22 @@ def w3_with_registry(w3):
...
+        b'safe-math-lib',  # pkg-name
         '1.0.0',  # version
-        'ipfs://QmbeVyFLSuEUxiXKwSsEjef6icpdTdA4kGG9BcrJXKNKUW'  # content-addressed URI
+        'ipfs://QmWgvM8yXGyHoGWqLFXvareJsoCZVsdrpKNCLMun3RaSJm'  # content-addressed URI
     ).transact()
     w3.eth.waitForTransactionReceipt(tx_hash)
     return w3, address, registry
 
 
 def test_package_init_from_registry(w3_with_registry):
```
njgheorghita: Same here, not sure how to avoid the http req - I'd like to use a real `valid_registry_uri` that is run through the validation utils - but also points to a valid IPFS content-addressed uri.

ethpm/backends/ipfs.py
```diff
@@ -0,0 +1,123 @@
...
+        raise NotImplementedError("Must be implemented by subclass.")
+
+
+class IPFSOverHTTPBackend(BaseIPFSBackend):
+    """
+    Base class for all IPFS URIs served over an http connection.
+    """
+    @property
+    @abstractmethod
+    def base_uri(self) -> str:
```
cburgdorf: No need to repeat the abstract method here as it is already in the `BaseIPFSBackend` just leave it out and it will implicitly continue to be the abstract method from the base class preventing anyone from direct instantiation of `IPFSOverHTTPBackend`

ethpm/package.py
```diff
@@ -1,11 +1,17 @@
+import json
```
cburgdorf: See comment above regarding imports

ethpm/utils/module_loading.py
```diff
@@ -0,0 +1,25 @@
+from importlib import import_module
+
+from typing import Any
+
+
+def import_string(dotted_path: str) -> Any:
```
cburgdorf: If that's a generic concept we reuse across multiple projects, maybe it deserves a place in `eth_utils`?

cburgdorf: (Not too relevant for this PR but maybe open up an issue to discuss and potentially move later)

tests/ethpm/backends/test_ipfs_backends.py
cburgdorf: Not sure what's available in Python land but maybe there's a lib where you can mock responses so that if a request is made to `http://somewhere.com/some/thing` you can intercept and send a response accordingly.

ethpm/backends/__init__.py
```diff
@@ -0,0 +1,27 @@
+import os
```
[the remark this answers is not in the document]
njgheorghita: Though - I just saw this `Standard library imports which import things from the module should be separated with a single line of whitespace.` - also do you tend to do multiline statements for a single import all the time? is there a reason?

ethpm/utils/module_loading.py
njgheorghita: sounds like a plan! i know it's also in py-evm, not sure about any other projects?

ethpm/backends/__init__.py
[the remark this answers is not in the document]
carver: Maybe just add `isort` and the configuration as they are in web3.py? Then you don't have to manually tweak or enforce the import sorting.

> do you tend to do multiline statements for a single import all the time? is there a reason?

`isort` takes care of this too. I'm fairly neutral on it, but it makes the diff nicer when going from one to two imports. (esp when adding trailing commas to cburgdorf's example)

ethpm/utils/module_loading.py
cburgdorf: Me neither but seems useful in general.

ethpm/backends/__init__.py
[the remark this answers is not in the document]
cburgdorf: @carver Agree but I would add that I'd be more and more in favor to just use `black` unless there are any big blockers (meaning non aesthetic things) keeping us from using it.

README.md
```diff
@@ -136,6 +137,15 @@ scheme://authority/package-name?version=x.x.x
 i.e. `ercxxx://packages.zeppelinos.eth/owned?version=1.0.0`
 
 
+## URI Backend Class
+* URI Backend Class defaults to `ethpm.backends.ipfs.IFPSGatewayBackend`
+* URI Backend Class is configurable via the environment variable `URI_BACKEND_CLASS` set to a dotted module path
```
pipermerriam: I'm of the opinion that these should be namespaced to prevent collisions (however unlikely).  (e.g. `ETHPM_URI_BACKEND_CLASS`)

ethpm/backends/__init__.py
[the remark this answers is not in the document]
pipermerriam: I'm also good with loosening those suggestions/requirements in the style guide as I think that specific suggestion is weirdly arbitrary.

ethpm/backends/base.py
```diff
@@ -0,0 +1,34 @@
...
+    @abstractmethod
+    def fetch_uri_contents(self, uri: str) -> bytes:
+        """
+        Fetch the contents stored at a URI.
+        """
+        raise NotImplementedError("Must be implemented by subclass.")
+
+    @property
+    @abstractmethod
+    def base_uri(self) -> str:
```
pipermerriam: Does this make sense as a required base method/property?  Seems like IPFS backend wouldn't use this. since they are just a scheme and then the hash `ipfs://Qm...`

This might belong in a second *base* class for `HTTP` based backends since they are likely to share similar logic.

ethpm/backends/ipfs.py
```diff
@@ -0,0 +1,101 @@
...
+        path = V2_PACKAGES_DIR / uri
+        return path.exists()
+
+    def base_uri(self) -> str:
+        pass
+
+
+class LocalIPFSBackend(BaseIPFSBackend):
+    """
+    Backend class for all IPFS URIs served through a locally running IPFS node.
```
pipermerriam: The word *locally* should probably be removed in favor of something like 

> served through a direct connection to an IPFS node

tests/ethpm/backends/test_ipfs_backends.py
pipermerriam: These are really integration style tests and their value seems questionable.  Maybe just 1-2 smoke tests (maybe using mocks) would be sufficient to know that everything is *wired up* correctly while still keeping the amount of mocking to a minimum.

tests/ethpm/test_package_init_from_ipfs.py
```diff
@@ -1,26 +1,18 @@
...
-    monkeypatch.setattr(ethpm.package, 'fetch_ipfs_package', mock_fetch)
-
 
 def test_package_from_ipfs_with_valid_uri():
-    package = Package.from_ipfs(VALID_IPFS_PKG)
+    os.environ['URI_BACKEND_CLASS'] = 'ethpm.backends.ipfs.DummyIPFSBackend'
+    package = Package.from_ipfs('safe-math-lib/1.0.0.json')
     assert package.name == 'safe-math-lib'
     assert isinstance(package, Package)
+    os.environ.pop('URI_BACKEND_CLASS')
```
pipermerriam: You should use [`monkeypatch.setenv`](https://docs.pytest.org/en/latest/monkeypatch.html) to ensure that things get cleaned up.

tests/ethpm/test_package_init_from_registry.py
pipermerriam: I'd suggest using a dummy backend which stores a local mapping of `URI -> content` and blows up if you ask it for anything it doesn't know about.

tests/ethpm/test_package_init_from_registry.py
```diff
@@ -40,20 +32,22 @@ def w3_with_registry(w3):
...
+    os.environ['URI_BACKEND_CLASS'] = 'ethpm.backends.ipfs.IPFSGatewayBackend'
     w3, address, registry = w3_with_registry
-    valid_registry_uri = 'ercXXX://%s/owned?version=1.0.0' % address
+    valid_registry_uri = 'ercXXX://%s/safe-math-lib?version=1.0.0' % address
     pkg = Package.from_registry(valid_registry_uri, w3)
     assert isinstance(pkg, Package)
-    assert pkg.name == 'owned'
+    assert pkg.name == 'safe-math-lib'
+    os.environ.pop('URI_BACKEND_CLASS')
```
pipermerriam: Same here with making use of the `monkeypatch` pytest fixture.

tests/ethpm/utils/test_uri_utils.py
```diff
@@ -15,10 +17,12 @@
     )
 )
 def test_get_manifest_from_content_addressed_uris_for_supported_schemes(uri, source):
+    os.environ['URI_BACKEND_CLASS'] = 'ethpm.backends.ipfs.DummyIPFSBackend'
     manifest = get_manifest_from_content_addressed_uri(uri)
     assert 'version' in manifest
     assert 'package_name' in manifest
     assert 'manifest_version' in manifest
+    os.environ.pop('URI_BACKEND_CLASS')
```
pipermerriam: Same here

ethpm/backends/base.py
njgheorghita: Maybe not. In the local IPFS node backend, I used this property to return `host:port` where the IPFS daemon is running. But I could see that maybe being confusing/ incorrectly named? And moving this property up a level to only `HTTP` based backends

ethpm/backends/__init__.py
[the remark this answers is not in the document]
njgheorghita: Thoughts on all single and double import statements being single lined? i.e.
```
from eth_utils import keccak
from eth_utils import keccak, to_text
```
This diverges from the style guide, but seems to be the only serious style conflict from using `black` to standardize formatting. Which also caught a fair amount of other style issues (i.e. only double quotes, formatting multiline args, etc.).

tests/ethpm/test_package_init_from_registry.py
njgheorghita: Thoughts on using the `requests-mock` library, like I did in a [subsequent pr](https://github.com/ethpm/py-ethpm/pull/60/files#diff-d76056684bd2b90a26b7829883ac9614R30). It's not ideal right now since it mocks ANY http request, I'm working on how to only mock a specific request and blow up for others (a little tricky since the configuration I'm using isn't the default one in the docs).

njgheorghita:
copying here since it got lost in a force push.
> I'd suggest using a dummy backend which stores a local mapping of URI -> content and blows up if you ask it for anything it doesn't know about.

I used the `requests-mock` library in a subsequent pr. Thoughts? It's not great yet since it stubs out ANY http request, but i'm working making it blow up if not making the right request (the docs for the specific configuration i'm using aren't well documented)

ethpm/backends/__init__.py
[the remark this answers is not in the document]
pipermerriam: >  I'm partial to deferring the style decisions to whatever `black` enforces, as long as it's consistent throughout the library.

👍

pipermerriam:
@njgheorghita 

I have strong opinions about mocks and I'll try to distill them here briefly.

## Mocks are to be avoided.

- Mock based tests **tend** to end up testing the implementation and not the actual functionality.  
- It is **very** easy to have mock-based tests pass but the actual code fail.
- can end up being *leaky* if not done correctly (leaving objects patched after test cleanup, or other weird side effects)

## Signal that your code is incorrectly structured

In many cases, finding that you need to mock something is a *signal* that you need to re-factor your code to make things more modular so you can test a specific component without needing to mock out other components.

## When all else fails

Mocks **can** be ok when all other options are exhausted, or it is the pragmatic solution, or some other reason to ignore the bad things.

## But maybe don't mock things

My opinion is that if you are going to mock the request/response cycle, why not just use a dummy backend and stub out the responses as fixtures.  Unless you **need** to specifically test the HTTP backends (which should be tested in isolation), then if you're testing other functionality, then mocking out the request/response cycle seems like un-necessary overhead where a generic fixture based dummy backend would be simpler.

njgheorghita:
@pipermerriam Thanks for writing those thoughts out, it was helpful.  

I changed the `Package.from_registry()` integration test from using a mock to the dummy backend. 
But it seems to me like a mock is still useful when unit testing the `fetch_uri_contents()` method on the IPFS gateway classes [here](https://github.com/ethpm/py-ethpm/pull/51/files#diff-d76056684bd2b90a26b7829883ac9614R31), since that's where the actual http req is made?

ethpm/backends/base.py
davesque: I think I agree with @pipermerriam here.  Seems like `base_uri` should be in a subclass for HTTP backends.

ethpm/backends/ipfs.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+
+    def can_handle_uri(self, uri: str) -> bool:
+        """
+        Return a bool indicating whether or not this backend
+        is capable of serving the content at the URI.
+        """
+        return is_ipfs_uri(uri)
+
+    def fetch_uri_contents(self, uri: str) -> bytes:
```
davesque: Since `fetch_uri_contents` is an abstract method in the `BaseURIBackend` class, I don't think you should need it here.

ethpm/backends/ipfs.py
```diff
@@ -0,0 +1,104 @@
...
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @property
+    def base_uri(self) -> str:
+        raise NotImplementedError("Must be implemented by subclass")
+
+
+class IPFSOverHTTPBackend(BaseIPFSBackend):
```
davesque: I think you can just put an abstract `base_uri` method on this class to take Piper's suggestion of having `base_uri` on an abstract class for HTTP backends.

README.md
```diff
@@ -136,6 +137,15 @@ scheme://authority/package-name?version=x.x.x
 i.e. `ercxxx://packages.zeppelinos.eth/owned?version=1.0.0`
 
 
+## URI Backend Class
```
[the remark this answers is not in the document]
njgheorghita: Word, i've gone ahead and done that in #58

ethpm/backends/ipfs.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def fetch_uri_contents(self, uri: str) -> bytes:
+        ipfs_hash = extract_ipfs_path_from_uri(uri)
+        gateway_uri = self.base_uri + ipfs_hash
+        response = requests.get(gateway_uri)
+        response.raise_for_status()
+        return response.content
+
+    @property
+    def base_uri(self) -> str:
```
davesque: This should be decorated with `abc.abstractmethod`.

PR #52, opened 2018-06-28
Isort

njgheorghita:
### What was wrong?
Import sorting was too subjective.

### How was it fixed?
Use `isort` library and config's from `web3` to sort import statements

#### Cute Animal Picture
image


davesque:
This looks good.  Did you by chance use the isort config from the project template?

njgheorghita:
@davesque Grabbed it from web3 - but looks to be just the same as the template. The only difference is that in web3, pytest is listed as a `known_standard_library` and in the template it's listed as a `known_third_party`, the template is correct, right?

njgheorghita:
Acutally, i'm not too sure - I'm starting to think i like sticking `pytest` in with the standard lib import statements

davesque:
The template is correct.  They had been doing that since `pytest` is so commonly used as to be almost like a standard library for us.  However, I suggested that this is really confusing to newcomers and it violates the principle of least astonishment.  So we updated the template to mark `pytest` as a known third party module, which it is.

njgheorghita:
@davesque good to know - i'll change it to `known_third_party`

davesque:
In general, it's good to look to the template whenever you're manually importing configuration stuff like that.  That will hopefully allow us to avoid inconsistent configs.  I sort of wish we had an official python project template actually instead of Jason's repo.

davesque:
Also, don't forget to re-run isort since it's going to change where that module shows up :).

davesque:
Schweet.  Looks like you rebuilt things?

njgheorghita:
Haha - yeah - and a typo in `tox.ini` had me tearing out my hair (or what's left of it) for a bit - but everything should be set now

davesque:
Can you merge or do you need me to?

njgheorghita:
I got it!

PR #60, opened 2018-07-02
Local ipfs backend

njgheorghita:
Branched from #51 

### What was wrong?
Connecting to a local ipfs node needed implementation

### How was it fixed?
Implemented `LocalIPFSBackend` Class

#### Cute Animal Picture
image


tests/ethpm/backends/test_ipfs_backends.py
```diff
@@ -27,6 +32,15 @@ def test_ipfs_and_infura_gateway_backends_fetch_uri_contents(
         assert contents_dict["package_name"] == "safe-math-lib"
 
 
+def test_local_ipfs_backend():
+    uri = "ipfs://Qme4otpS88NV8yQi8TfTP89EsQC5bko3F5N1yhRoi6cwGV"
+    backend = LocalIPFSBackend("localhost", "8080")
+    with requests_mock.Mocker() as m:
```
njgheorghita: I might be mistaken, but this seems like an appropriate place to use the mocker to me, since it's testing the `fetch_uri_contents` method which makes an req unique to the `LocalIPFSBackend` class. Using the `DummyIPFSBackend` here makes less sense to me

davesque: We've already discussed this so I'm just leaving this comment here for the record.  It seems better to change the `LocalIPFSBackend` class to accept an already instantiated `Client` instance from the `ipfsapi` library as the only argument to `__init__` which would get stored in `self.client`.  Then, the `fetch_uri_contents` method would just do  `self.client.cat`.  That would make it easier to inject fake dependencies for that backend for the purpose of testing.  This approach is advocated in this article written by the principal engineer at SendGrid: https://sendgrid.com/blog/when-writing-unit-tests-dont-use-mocks/ .

tests/ethpm/backends/test_ipfs_backends.py
```diff
@@ -27,6 +33,17 @@ def test_ipfs_and_infura_gateway_backends_fetch_uri_contents(
...
 
+def test_local_ipfs_backend(monkeypatch):
+    uri = "ipfs://Qme4otpS88NV8yQi8TfTP89EsQC5bko3F5N1yhRoi6cwGV"
+    ipfs_client = ipfsapi.Client("localhost", "8080")
+    def mockreturn(path):
+        return b"pragma"
+    monkeypatch.setattr(ipfs_client, 'cat', mockreturn)
+    backend = LocalIPFSBackend(ipfs_client)
+    contents = backend.fetch_uri_contents(uri)
+    assert contents.startswith(b"pragma")
```
njgheorghita: @davesque Is this something like what you had in mind?

njgheorghita:
@davesque Just thinking out loud - but what are the consequences of having an `init` in the base URI class (regardless of whether it's initiated with (`host`&`port` or a `Client`). 
Backends are currently set as an env variable - with the idea (i think) being that you can switch out the used backend with ease simply by changing the env variable. But none of the other backends have an `init` and I'm not quite sure if this will cause some hiccups / extra setup if you want to use the local ipfs backend?

tests/ethpm/backends/test_ipfs_backends.py
```diff
@@ -27,6 +33,19 @@ def test_ipfs_and_infura_gateway_backends_fetch_uri_contents(
         assert contents_dict["package_name"] == "safe-math-lib"
 
 
+def test_local_ipfs_backend(monkeypatch):
+    uri = "ipfs://Qme4otpS88NV8yQi8TfTP89EsQC5bko3F5N1yhRoi6cwGV"
+    ipfs_client = ipfsapi.Client("localhost", "8080")
+
+    def mockreturn(path):
```
davesque: Yeah, that's one way to do what I was thinking.  More specifically, I had imagine something like this which seems a bit more in line with what was discussed in the article:
```python
class FakeClient:
    def cat(self, ipfs_hash):
        return b"pragma"

backend = LocalIPFSBAckend(FakeClient())
...
```
What you did is effectively the same but just a bit more complicated.

You could also maybe do something like this which would also verify that the correct argument was given to `cat`:
```python
class FakeClient:
    def cat(self, ipfs_hash):
        return (ipfs_hash + "_contents").encode("utf-8")

backend = LocalIPFSBAckend(FakeClient())
...
assert contents == b"Qme4otp..._contents"
```

davesque:
@njgheorghita As per your concerns about other backends not having an `__init__` method (which takes arguments), you could change the `LocalIPFSBackend` to something like this:
```python
class LocalIPFSBackend(...):
    def __init__(self):
        self.client = ipfsapi.Client(...)
```
Then, in your tests, you could do something like this:
```python
backend = LocalIPFSBackend()
backend.client = FakeClient()
...
```

njgheorghita:
@davesque Made the change to use a `FakeClient` class instead of monkeypatching. Thanks for the tip. 
In terms of the backend, I'm unsure about using `get_uri_backend` in our code, not necessarily the tests. For example [here](https://github.com/ethpm/py-ethpm/blob/master/ethpm/utils/uri.py#L23) - I'm thinking we want a way to pass in the `host` and `port` as args somewhere if `ETHPM_URI_BACKEND` is set to `LocalIPFSBackend`. Or we can just set default `host=localhost` and `port=5001` (default for ipfs daemon) and then an additional function that would allow the user to change those params

davesque:
You can also just make a subclass of `LocalIPFSBackend` that does something different in the `__init__` method.  Then you can still have backends that take no `__init__` args.

njgheorghita:
@davesque Made the changes we talked about 
- env variable `ETHPM_URI_BACKEND_CLASS` => `ETHPM_IPFS_BACKEND_CLASS`
- default to IPFS r&w port (punt configure a separate port via another env variable down the road, or until it seems like it would prove useful to someone)
- lazily raise `ipfsapi` connection errors

I think that should cover this pr - tho let me know if i missed something

PR #61, opened 2018-07-02
Http backend

njgheorghita:
### What was wrong?
URI backend needed to serve content-addressed github URIs

### How was it fixed?
Wrote backend that serves github URIs conforming to 

`https://raw.githubusercontent.com/user/repo/commit_hash/path/to/manifest.json`

#### Cute Animal Picture
image


njgheorghita:
@pipermerriam ping for review. The scheme i'm using here isn't quite exactly, what's discussed in the ERC 1319](https://github.com/ethereum/EIPs/issues/1319), I'm not sure if that's a problem or not. IMO just using the `raw.githubusercontent.com` authority is simpler than creating/fetching blobs as suggested in the spec. Any thoughts? Obvs, i'd like to conform to the spec wherever possible, but this is also an area that imo doesn't require strict conformity. 

I also thought about requiring a content hash fragment suffixing the github uri . . . i.e.
`https://raw.githubusercontent.com/ethpm/ethpm-spec/481739f6138907db88602558711e9d3c1301c269/examples/owned/contracts/Owned.sol#bfdea1fa5f33c30fee8443c5ffa1020027f8813e0007bb6f82aaa2843a7fdd60` 
for an added layer of security, but I recently removed that since the commit-hash is already in the uri, didn't seem necessary to have two content hashes

pipermerriam:
@njgheorghita on not including the hash in the URL: I don't think this is ok.  While I agree that it doesn't allow for arbitrary modification of the resouce by the user, it doesn't provide protection against github serving up incorrect content.  I acknowledge this probably isn't a concern we're likely to see realized, but I'd prefer we do it *correctly* and both require the hash as well as verifying the payload we receive matches said hash.

Which makes me realize that we should be doing the same with the IPFS gateway backends (and probably generically via any of the IPFS backends) just to be doubly sure that the content we receive is indeed the content that should have been at the address.

njgheorghita:
@pipermerriam Makes sense, added the requirement for a valid content hash to the end of github uris. And added a validation to IPFS backend that the hash of the retrieved contents matches the content hash in the uri

docs/uri_backends.rst
```diff
@@ -1,31 +1,63 @@
...
+.. code:: python
+
+   https://raw.githubusercontent.com/user/repo/commit_hash/path/to/manifest.json#content_hash
+
+To generate a valid Github PM URI.
+
+- Go to the target manifest in your browser.
+- Press ``y`` to generate the permalink in the address bar.
+- Replace ``"github"`` with ``"raw.githubusercontent"``, and remove the ``"blob"`` namespace from the URI.
+- Suffix the URI with ``#`` followed by the ``keccak`` hash of the bytes found at the Github URI.
```
pipermerriam: Not for this PR, but it'd be pretty cool to provide a utility on the `GithubOverHTTPSBackend` that did this automatically with a uri and supported a handful of formats.

```python
>>> GithubOverHTTPSBackend.get_permalink_with_content_hash('https://github.com/ethpm/ethpm-spec/blob/master/examples/owned/1.0.0.json')
"https://raw.githubusercontent.com/ethpm/ethpm-spec/3945c47dedb04930ee12c0281494a1b5bdd692a0/examples/owned/1.0.0.json#0x01cbc2a69a9f86e9d9e7b87475e2ba2619404dc8d6ee3cb3a8acf3176c2cace1"
```

Should be able to:

* parse the repository name and file path and branch
* use the github API to pull the latest commit hash
* generate the `raw.github.com` URL
* pull the file and compute the `keccak` of the contents.

This would be something well defined enough that it could be bountied.

ethpm/backends/http.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def can_translate_uri(self, uri: str) -> bool:
+        """
+        GithubOverHTTPSBackend uri's must resolve to a valid manifest,
+        and cannot translate to another content-addressed URI.
+        """
+        return False
+
+    def fetch_uri_contents(self, uri: str) -> bytes:
+        http_uri, validation_hash = uri.split("#")
```
pipermerriam: You should use `urllib.parse.urlpase` and pull the `fragment` from that return value.

ethpm/backends/ipfs.py
```diff
@@ -56,7 +61,14 @@ def __init__(self) -> None:
 
     def fetch_uri_contents(self, uri: str) -> bytes:
         ipfs_hash = extract_ipfs_path_from_uri(uri)
-        return self.client.cat(ipfs_hash)
+        contents = self.client.cat(ipfs_hash)
+        validation_hash = generate_file_hash(contents)
+        if validation_hash != ipfs_hash:
+            raise ValidationError(
+                "Hashed IPFS contents retrieved from uri: {0} "
```
pipermerriam: If you're python 3.6+ you can start using f-strings for formatting: https://realpython.com/python-f-strings/

ethpm/validation.py
```diff
@@ -167,3 +174,18 @@ def validate_single_matching_uri(all_blockchain_uris: List[str], w3: Web3) -> st
...
+    """
+    Validate that the keccak(contents) matches the validation_hash.
+    """
+    hashed_contents = keccak(contents)
+    decoded_validation = decode_hex(validation_hash)
+    if hashed_contents != decoded_validation:
+        raise ValidationError(
+            "Invalid content-addressed URI. "
+            "Validation hash:{0} does not match the hash of URI contents: {1}.".format(
+                decoded_validation, hashed_contents
```
pipermerriam: Probably good to output these in their hex encoded formats for friendlier error messages.

tests/conftest.py
```diff
@@ -104,6 +104,12 @@ def _get_factory(package, factory_name):
     return _get_factory
 
 
+@pytest.fixture
+def owned_contract():
+    with open(V2_PACKAGES_DIR / "owned" / "contracts" / "Owned.sol") as file_obj:
```
pipermerriam: I think you can also do `(V2_PACKAGES_DIR / "owned" / "contracts" / "Owned.sol").open()` but I'm undecided on whether I think it's a good pattern.  The outer parenthesis wrapping is sort of awkward.  Not advocating change, just bringing up my thoughts.

tests/ethpm/test_dependencies.py
```diff
@@ -43,7 +43,7 @@ def test_get_dependency_package(dependencies):
 
 def test_validate_build_dependencies(dummy_ipfs_backend, piper_coin_manifest):
     result = validate_build_dependency(
-        "standard-token", "ipfs://QmVu9zuza5mkJwwcFdh2SXBugm1oSgZVuEKkph9XLsbUwg"
+        "standard-token", "ipfs://QmVu9zuza5mkJwwcFdh2SXBugm1oSgZVuEKkph9XLsbUwg#0x123"
```
pipermerriam: What's this doing?

docs/uri_backends.rst
```diff
@@ -1,31 +1,63 @@
...
 ========================
 
+BaseURIBackend
+--------------
+
+``Py-EthPM`` uses the ``BaseURIBackend`` as the parent class for all of its URI backends. To write your own backend, it must implement the following methods. 
+
+.. py:method:: BaseURIBackend.can_resolve_uri(uri)
+
+   Returns a bool indicating whether or not this backend is capable of resolving the given URI to a manifest.
```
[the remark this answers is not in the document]
njgheorghita: 👍 autodoc, moving all (or as many) of the docs over to autodoc is next on my todo list

tests/conftest.py
njgheorghita: Agreed on the parenthesis wrapping being weird, but i'm a fan of pathlib and think the weirdness is worth it

tests/ethpm/test_dependencies.py
njgheorghita: 😬 nothing 😬

PR #64, opened 2018-07-06
Improve error message when schema invalidates

njgheorghita:
### What was wrong?
Error message for invalidated manifests was lacking
Also had to update `generate_contract_factory_kwargs` to use the new bytecode object field definition of `deployment_bytecode`

### How was it fixed?
Updated message to include the reason for what went wrong during validation

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/package.py
+++ b/ethpm/package.py
@@ -72,8 +72,8 @@ class Package(object):
             validate_minimal_contract_factory_data(contract_data)
         except KeyError:
             raise InsufficientAssetsError(
-                "This package has insufficient package data to generate"
-                "a contract_factory for contract: {0}.".format(name)
+                "This package has insufficient package data to generate "
+                "a contract factory for contract:{0}.".format(name)
             )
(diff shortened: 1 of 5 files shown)


ethpm/package.py
```diff
@@ -72,7 +72,7 @@ def get_contract_factory(self, name: ContractName, w3: Web3 = None) -> Contract:
             validate_minimal_contract_factory_data(contract_data)
         except KeyError:
             raise InsufficientAssetsError(
-                "This package has insufficient package data to generate"
+                "This package has insufficient package data to generate "
                 "a contract_factory for contract: {0}.".format(name)
```
davesque: This is not related to the changes you made, but I wonder if you could change "contract_factory" here to "contract factory".  The underscore seems kind of random.

njgheorghita: Made the change - feel free to hit merge if everything else looks good

PR #65, opened 2018-07-07
Memoize get backend and update IPFS gateway URI

davesque:
### What was wrong?

We are instantiating URI backends every time URIs are requested.  This doesn't seem quite right.  Also, the IPFS gateway URL was outdated.

### How was it fixed?

Added `lru_cache` decorator to `get_uri_backend` function.  Also updated IPFS gateway URL.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/ethpm/backends/__init__.py
+++ b/ethpm/backends/__init__.py
@@ -1,3 +1,4 @@
+import functools
 import os
 
 from typing import Type
@@ -9,6 +10,7 @@ from ethpm.utils.module_loading import import_string
 DEFAULT_URI_BACKEND = "ethpm.backends.ipfs.IPFSGatewayBackend"
 
 
+@functools.lru_cache(maxsize=None)
 def get_uri_backend(import_path: str = None) -> BaseURIBackend:
     """
     Return the `BaseURIBackend` class specified by import_path, default, or env variable.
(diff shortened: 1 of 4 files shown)


davesque:
I guess we can take this or leave it.  Just seems to me like we shouldn't be created new backend instances every time a URI's contents are downloaded.

njgheorghita:
My only reservation kinda depends on the comment I just posted in #60 . How are we going to handle `LocalIPFSBackend` say if somebody wants to change the port from read-only (8080) to read&write (5001). Though - I guess we can just clear the cache when somebody calls the method that updates the `host` & `port` for `LocalIPFSBackend`?

davesque:
Maybe you could just have a subclass of `LocalIPFSBackend` called `LocalReadWriteIPFSBackend` and override the `__init__` method to pass in a different port.  Then the `ETHPM_URI_BACKEND` env could be set to `LocalReadWriteIPFSBackend`.

njgheorghita:
@davesque Let me know how you want to treat this pr. In #60 I removed `get_uri_backend` and replaced with `get_ipfs_backend` as we talked about. I can add the cache onto `get_ipfs_backend` in #60 or if you want to update this pr to reflect those changes, either way works for me

PR #67, opened 2018-07-09
Implement `Package.get_contract_instance`

njgheorghita:
### What was wrong?
Since we have `Package.get_contract_factory` it seemed useful to implement `Package.get_contract_instance`. So you can easily fetch an instance of a contract at an address (though the `package_data` MUST contain the abi for this contract in its `contract_types`).
ie. with
```python
token_package = web3.pm.get_package_from_manifest(standard_token_manifest)
ERC20 = token_package.get_contract_factory('StandardToken')
tx_hash = ERC20.constructor(100).transact()
tx_receipt = web3.eth.getTransactionReceipt(tx_hash)
address = tx_receipt["contractAddress"]
```
you can directly get the contract instance from the package
```python
erc20 = token_package.get_contract_instance('StandardToken', address)
total_supply = erc20.functions.totalSupply().call()
```
rather than using web3
```python
erc20 = web3.eth.contract(address=address, abi=ERC20.abi)
total_supply = erc20.functions.totalSupply().call()
```

### How was it fixed?
Wrote `Package.get_contract_instance` that uses similar validation to `get_contract_factory`

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/package.py
+++ b/ethpm/package.py
@@ -1,6 +1,7 @@
 import json
 from typing import Any, Dict
 
+from eth_typing import Address
 from eth_utils import to_text
 from web3 import Web3
 from web3.eth import Contract
@@ -57,15 +58,8 @@ class Package(object):
         """
         API to generate a contract factory class.
         """
-        current_w3 = None
-
-        if w3 is not None:
-            current_w3 = w3
-        else:
-            current_w3 = self.w3
-
         validate_contract_name(name)
-        validate_w3_instance(current_w3)
+        current_w3 = self.get_w3_instance(w3)
 
         try:
             contract_data = self.package_data["contract_types"][name]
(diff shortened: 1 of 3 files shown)


tests/ethpm/test_package.py
```diff
@@ -62,6 +71,21 @@ def test_get_contract_factory_throws_if_name_isnt_present(safe_math_package, w3)
         assert safe_math_package.get_contract_factory("DoesNotExist", w3)
 
 
+def test_get_contract_instance(deployed_safe_math):
+    safe_math_package, address = deployed_safe_math
+    contract_instance = safe_math_package.get_contract_instance("SafeMathLib", address)
+    assert contract_instance.address == address
```
pipermerriam: Generally best to use `eth_utils.is_same_address` in cases like this since one of these addresses could end up being checksummed (automatically) at some point at which this would break.

ethpm/package.py
```diff
@@ -80,6 +74,37 @@ def get_contract_factory(self, name: ContractName, w3: Web3 = None) -> Contract:
...
+        current_w3 = self.get_w3_instance(w3)
+
+        try:
+            abi = self.package_data["contract_types"][name]["abi"]
+        except KeyError:
+            raise InsufficientAssetsError(
+                "Package does not have the ABI required to generate a contract instance "
+                "for contract: {0} at address: {1}.".format(name, address)
+            )
+        contract_instance = current_w3.eth.contract(address=address, abi=abi)
```
pipermerriam: It would be good to include the `binary` and `runtime_binary` in the constructor kwargs if they are available in the provided package assets.

ethpm/package.py
```diff
@@ -80,6 +74,40 @@ def get_contract_factory(self, name: ContractName, w3: Web3 = None) -> Contract:
         contract_factory = current_w3.eth.contract(**contract_kwargs)
         return contract_factory
 
+    def get_contract_instance(
+        self, name: ContractName, address: Address, w3: Web3 = None
```
davesque: I'm forgetting, why do we accept `w3` as an argument to all of these methods?  Why don't we just always use `self.w3`?

njgheorghita: @davesque Well, we don't require that a `web3` when instantiating a `Package`. We provide a `Package.set_default_w3` to set the `web3` after creating a `Package`. @pipermerriam correct me if i'm wrong, but I think we want to leave the ability to pass in a `web3` to `Package.get_contract_factory` and `Package.get_contract_instance` - as it provides flexibility for the user to switch between chains - rather than strictly using the default `w3` set on `Package` for all web3 tasks.

pipermerriam: > I think we want to leave the ability to pass in a w3 to `Package.get_contract_factory` and `Package.get_contract_instance` - as it provides flexibility for the user to switch between chains

You are correct that when we originally talked about this that was the plan, however, I don't think it was the right choice.  It's probably easier for us to write the code and easier for consumers of the API to just require a web3 instance up front and then to not have to worry about it from there on.

PR #69, opened 2018-07-12
Support Packages with build dependencies

njgheorghita:
### What was wrong?
We want to be able to support packages with build dependencies.

### How was it fixed?
Wrote a `Dependencies` class that stores (`pkg_name` => `Package`) of the `"build_dependencies"` in a `Package`s manifest. Also performs validation to make sure `pkg_name` is a valid package name and the uri it's pointing to in the manifest is a valid IPFS uri (i.e. `ipfs://Qme2.....`)

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/dependencies.py
+++ b/ethpm/dependencies.py
@@ -1,0 +1,41 @@
+from typing import Dict, List
+
+
+class Dependencies:
+    """
+    Access build dependencies belonging to a package.
+    """
+
+    def __init__(self, build_dependencies: Dict[str, "Package"]) -> None:
+        self.build_dependencies = build_dependencies
+
+    def __getitem__(self, key: str) -> "Package":
+        return self.get(key)
+
+    def __contains__(self, key: str) -> bool:
+        return key in self.build_dependencies
+
+    def _validate_name(self, name: str) -> None:
+        if name not in self.build_dependencies:
+            raise KeyError(
+                "Package name: {0} not found in build dependencies.".format(name)
+            )
+
+    def get(self, key: str) -> "Package":
+        self._validate_name(key)
+        return self.build_dependencies.get(key)
+
+    def items(self) -> Dict[str, "Package"]:
+        item_dict = {name: self.get(name) for name in self.build_dependencies}
+        return item_dict
+
+    def values(self) -> List["Package"]:
+        values = [self.get(name) for name in self.build_dependencies]
+        return values
+
+    def get_dependency_package(self, package_name: str) -> "Package":
+        """
+        Return a the dependency Package of the given package name.
+        """
+        self._validate_name(package_name)
+        return self.get(package_name)
--- a/ethpm/dependencies.py
+++ b/ethpm/dependencies.py
@@ -1,41 +1,52 @@
-from typing import Dict, List
+from typing import Dict, List, Tuple
+
+from ethpm.validation import validate_package_name
 
 
 class Dependencies:
     """
-    Access build dependencies belonging to a package.
+    Class to manage the `Package` instances of a Package's `build_dependencies`.
     """
 
-    def __init__(self, build_dependencies: Dict[str, "Package"]) -> None:
+    # ignoring Package type here and below to avoid a circular dependency
+    def __init__(
+        self, build_dependencies: Dict[str, "Package"]  # type: ignore
+    ) -> None:
         self.build_dependencies = build_dependencies
 
-    def __getitem__(self, key: str) -> "Package":
-        return self.get(key)
+    def __getitem__(self, key: str) -> "Package":  # type: ignore
+        return self.build_dependencies.get(key)
 
     def __contains__(self, key: str) -> bool:
         return key in self.build_dependencies
 
     def _validate_name(self, name: str) -> None:
+        validate_package_name(name)
         if name not in self.build_dependencies:
             raise KeyError(
                 "Package name: {0} not found in build dependencies.".format(name)
             )
 
-    def get(self, key: str) -> "Package":
-        self._validate_name(key)
-        return self.build_dependencies.get(key)
+    def items(self) -> Tuple[Tuple[str, "Package"], ...]:  # type: ignore
+        """
+        Return an iterable containing package name and
+        corresponding `Package` instance that are available.
+        """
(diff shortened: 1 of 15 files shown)


ethpm/dependencies.py
```diff
@@ -0,0 +1,41 @@
...
+    def __contains__(self, key: str) -> bool:
+        return key in self.build_dependencies
+
+    def _validate_name(self, name: str) -> None:
+        if name not in self.build_dependencies:
+            raise KeyError(
+                "Package name: {0} not found in build dependencies.".format(name)
+            )
+
+    def get(self, key: str) -> "Package":
```
pipermerriam: The object you've created here closely maps to a dictionary, but the behavior of this method would be surprising as it raises a `KeyError` if the dependency is not present.  Normal dictionaries just return `None` on a call to `get` with a missing key.

I think the most appropriate thing to do would be dropping the `.get` method since `get_dependency_package` holds more semantic meaning and performs the same task.

ethpm/package.py
```diff
@@ -142,6 +142,26 @@ def name(self) -> str:
...
+        """
+        Retrieve instance of `Dependencies` (i.e. build dependencies)
+        belonging to this Package.
+        """
+        validate_build_dependencies_are_present(self.package_data)
+
+        dependencies = self.package_data["build_dependencies"]
+        validate_build_dependencies(dependencies)
+        dependency_packages = {
+            dependency: Package.from_ipfs(uri)
```
pipermerriam: I think we'll want to be a bit more fancy here and catch any failures in this `Package.from_ipfs(uri)` in such a way that we can raise an error with more useful information.

Something like:

> Failed to retrieve build dependency `safe-math-lib` from URI `ipfs://Qm....`: Got error: <the-error>

ethpm/validation.py
```diff
@@ -17,6 +19,16 @@ def validate_package_name(pkg_name: str) -> None:
...
 
 
+def validate_build_dependencies(dependencies: Dict[str, str]) -> None:
+    """
+    Raise an exception if the key in dependencies is not a valid package name,
+    or if the value is not a valid IPFS URI.
+    """
+    for key, value in dependencies.items():
+        validate_package_name(key)
+        is_ipfs_uri(value)
```
pipermerriam: This function just returns `True/False`.  I'd suggest creating a new `validate_ipfs_URI` function:

```python
def validate_ipfs_URI(value):
    if not is_ipfs_uri(value):
        raise ValidationError(...)
```

tests/ethpm/test_dependencies.py
```diff
@@ -0,0 +1,65 @@
...
+from ethpm import Package
+from ethpm.dependencies import Dependencies
+from ethpm.exceptions import UriNotSupportedError
+from ethpm.validation import validate_build_dependencies
+
+
+@pytest.fixture
+def dependencies(monkeypatch, standard_token_manifest, piper_coin_manifest):
+    monkeypatch.setenv(
+        "ETHPM_IPFS_BACKEND_CLASS", "ethpm.backends.ipfs.DummyIPFSBackend"
```
pipermerriam: You might make a fixture that does this so that you can easily re-use it.

```python
@pytest.fixture
def dummy_ipfs_backend(monkeypatch):
    ... # do the monkeypatch here

@pytest.fixture
def dependencies(dummy_ipfs_backend, ...):
    # anything that uses this dependency will also automatically monkeypatch the ipfs backend)
    ...
```

tests/conftest.py
```diff
@@ -51,6 +51,18 @@ def owned_manifest():
         return json.load(file_obj)
 
 
+@pytest.fixture
+def piper_coin_manifest():
+    with open(str(V2_PACKAGES_DIR / "piper-coin" / "1.0.0-pretty.json")) as file_obj:
+        return json.load(file_obj)
```
pipermerriam: Not a big deal, but an optimization you could do for these.

```
@pytest.fixture(scope="session")
def _piper_coin_manifest():
    with open(str(V2_PACKAGES_DIR / "piper-coin" / "1.0.0-pretty.json")) as file_obj:
        return json.load(file_obj)

@pytest.fixture
def piper_coin_manifest(_piper_coin_manifest):
    return copy.deepcopy(_piper_coin_manifest)
```

This way we only every hit the disk once per test run.  Alternatively, you could just load the manifest as a top level module variable instead of the `_piper_coin_manifest` fixture, and just return a copy of that variable.

tests/ethpm/test_dependencies.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@pytest.fixture
+def invalid_manifest(piper_coin_manifest):
+    piper_coin_manifest["build_dependencies"]["_invalid_package_name"] = []
+    return piper_coin_manifest
+
+
+def test_dependencies_implements_getitem(dependencies, safe_math_lib_package):
+    assert isinstance(dependencies, Dependencies)
```
pipermerriam: I think we can either drop this assertion **or** move it into the `dependencies` fixture itself if you want to maintain the sanity check.

ethpm/dependencies.py
```diff
@@ -0,0 +1,41 @@
...
+        self._validate_name(key)
+        return self.build_dependencies.get(key)
+
+    def items(self) -> Dict[str, "Package"]:
```
pipermerriam: This is also odd.  The `items` method normally returns an iterable of 2-tuples.

```python
In [76]: x = {'a': 1, 'b': 2, 'c': 3}

In [77]: tuple(x.items())
Out[77]: (('a', 1), ('b', 2), ('c', 3))

```

njgheorghita:
@davesque Totally fair. I started with the dict approach then decided to move to the class approach b/c 1. it follows the approach we used for `Deployments` and in my head `deployments` and `dependencies` are fairly similar approaches. 2. it's more flexible/extensible. While it's hard for me to say 100% that we'll need this flexibility down the road, it's possible. For instance I just added `validate_package_name` to `Dependencies._validate_name` fairly easily. 

Again, I can't say 100% that the benefits above outweigh the introduced overhead - but if you think the dict approach is the better option, i'm open to moving it back.

PR #75, opened 2018-07-17
Validate raw manifests are formatted correctly

### LINKED ISSUE #71: Force Packages to be created with compact manifests.
### What was wrong?
Valid manifests must be packed according to specific rules. 
```
The document must be tightly packed, meaning no linebreaks or extra whitespace.
The keys in all objects must be sorted alphabetically.
Duplicate keys in the same object are invalid.
The document must use UTF-8 encoding.
The document must not have a trailing newline.
```
We need to enforce that manifests used to instantiate packages follow these rules

### How can it be fixed?

Fill this section in if you know how this could or should be fixed.

njgheorghita:
### What was wrong?
The ethpm-spec requires certain constraints on how a manifest is formatted . . . #71 

### How was it fixed?
Added a validation util `validate_raw_manifest_format` to class methods for instantiating a package from a file

#### Cute Animal Picture
image

fixes #71


njgheorghita:
This pr validates manifests loaded from file. Yet, when the same check isn't really useful when instantiating a `Package` straight from a dict object. Do we want to perform any validation for that? I.e. alphabetically sorted keys (i'm not sure if the other constraints (duplicate keys or extra newlines) even apply in this case, since we're dealing with a dict object 🤷‍♂️ )

njgheorghita:
@piper - ping!

njgheorghita:
> I'm 👎 on not requiring a content hash.

@pipermerriam  It may have been lost to time/force pushes, but do you remember the context for this? I'm not really sure where requiring a content hash would play into this pr (maybe referring to the hash of a manifest's contents?) - in which case requiring that hash seems excessive to me 🤷‍♂️ Otherwise this is ready for 👓

tests/ethpm/utils/test_manifest_validation_utils.py
```diff
@@ -7,10 +7,66 @@
...
     validate_manifest_deployments,
     validate_manifest_exists,
+    validate_raw_manifest_format,
 )
 from ethpm.validation import validate_manifest_version
 
 
+@pytest.fixture
+def all_raw_strict_manifests(package_names):
+    manifests = []
```
pipermerriam: Just my general dislike for mutation.  Converting this to use `to_tuple` and using a `yield` pattern should work here.

tests/ethpm/utils/test_manifest_validation_utils.py
```diff
@@ -7,10 +7,66 @@
...
+    return manifests
+
+
+@pytest.fixture
+def all_pretty_manifests(package_names):
+    manifests = []
+    for name in package_names:
+        with open(str(V2_PACKAGES_DIR / name / "1.0.0-pretty.json")) as file_obj:
+            manifests.append(file_obj.read())
+    return manifests
```
pipermerriam: Instead of having the fixture be a list of manifests, I think it would be better to parametrize the fixture which I believe also cleans up the tests below.

```python
@pytest.fixture.params(ALL_PACKAGE_NAMES)
def pretty_manifest(request):
    with open(str(... / request.param / "1.0.0-pretty.json")) as file_obj:
        return file_obj.read()
```

Then in your tests, if you use the `pretty_manifest` fixture, it will run the test once for each package name.

PR #76, opened 2018-07-17
Require w3 instance upon package instantiation

njgheorghita:
### What was wrong?
Optional `w3` upon `package` instantiation became unwieldy - a la discussion in #67 

### How was it fixed?
Instead of making a `w3` instance optional upon `Package` instantiation, require it whenever creating a Package object

- I left `Package.set_default_w3` so that you can change the default `w3` instance - which seems like something we want? (ie. to switch chains to fetch a certain deployment??)

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/package.py
+++ b/ethpm/package.py
@@ -36,12 +36,11 @@ from ethpm.validation import validate_build_dependency, validate_registry_uri
 
 
 class Package(object):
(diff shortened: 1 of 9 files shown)


njgheorghita:
Closing since #80 was merged (since it was approved) that contained these changes

PR #77, opened 2018-07-18
Replace standalone manifests only in testing with all manifests

njgheorghita:
### What was wrong?
Some tests were only using standalone manifests

### How was it fixed?
Changed fixture from `all_standalone_manifests` to `all_manifests`

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/utils/manifest_validation.py
+++ b/ethpm/utils/manifest_validation.py
@@ -1,7 +1,6 @@
-import itertools
 import json
 import os
-from typing import Any, Dict
+from typing import Any, Dict, List, Set
 
 from jsonschema import ValidationError as jsonValidationError, validate
 
@@ -14,6 +13,15 @@ MANIFEST_SCHEMA_PATH = str(SPEC_DIR / "package.spec.json")
 def _load_schema_data() -> Dict[str, Any]:
     with open(MANIFEST_SCHEMA_PATH) as schema:
         return json.load(schema)
+
+
+def extract_contract_types_from_deployments(deployment_data: List[Any]) -> Set[str]:
+    contract_types = set(
+        deployment["contract_type"]
+        for chain_deployments in deployment_data
+        for deployment in chain_deployments.values()
+    )
+    return contract_types
 
 
 def validate_manifest_against_schema(manifest: Dict[str, Any]) -> None:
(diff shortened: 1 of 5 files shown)


ethpm/utils/manifest_validation.py
```diff
@@ -54,11 +54,11 @@ def validate_manifest_deployments(manifest: Dict[str, Any]) -> None:
     if set(("contract_types", "deployments")).issubset(manifest):
         all_contract_types = list(manifest["contract_types"].keys())
         all_deployments = list(manifest["deployments"].values())
-        all_deployment_names = set(
-            itertools.chain.from_iterable(deployment for deployment in all_deployments)
-        )
+        all_deployment_contract_types = [
+            cytoolz.first(dep.values())["contract_type"] for dep in all_deployments
+        ]
```
njgheorghita: This was incorrect as previously we were using the deployment dictionary key (contract_name) to compare against `all_contract_types` instead of the 'contract_type' key inside the deployment.
However, i'm not sure if this approach is too complex / unreadable? Open to any suggestions on improving the readability

pipermerriam: I'm actually not sure this approach is correct.  The `cytoolz.first` call seems like it would truncate the deployment contracts for a manifest that has multiple deployments.  It seems like this should be

```
all_deployment_names = set(
    deployment['contract_type'] for chain_deployments in all_deployments
    for deployment in chain_deployments.values()
)
```

This is probably worth writing a simple utility `extract_contract_types_from_deployments` so that we can write some quick tests for it.

tests/ethpm/test_package.py
```diff
@@ -20,14 +20,14 @@ def deployed_safe_math(safe_math_package, w3):
     return safe_math_package, tx_receipt.contractAddress
 
 
-def test_package_object_instantiates_with_a_web3_object(all_standalone_manifests, w3):
-    for manifest in all_standalone_manifests:
+def test_package_object_instantiates_with_a_web3_object(all_manifests, w3):
+    for manifest in all_manifests:
```
pipermerriam: we often try to avoid loops like this in tests in favor of either:

- parametrized fixtures
- parametrized tests

```python
@pytest.fixture(params=[1, 2, 3])
def fixture_with_params(request):
    return request.param

@pytest.mark.parmetrize('value', [1, 2, 3])
def test_that_is_parametrized(value):
    ...
```

PR #79, opened 2018-07-24
Generalize URI handling (`Package.from_uri`)

### LINKED ISSUE #37: Add configurable IPFS URI backend

njgheorghita:
### What was wrong?
A more generalized scheme for handling package instantiation from URI was needed.

### How was it fixed?
Got rid of `Package.from_ipfs` / `Package.from_registry` in favor of `Package.from_uri` which will return a Package object instantiated by a manifest found at a valid content-addressed URI. 

#### Cute Animal Picture
image

fixes #37


njgheorghita:
@pipermerriam ping when you get the chance

ethpm/exceptions.py
```diff
@@ -23,6 +23,14 @@ class ValidationError(PyEthPMError):
...
+    """
+    Raised when the given URI cannot be served by any of the available backends.
+    """
+
+    pass
+
+
 class UriNotSupportedError(ValidationError):
     """
     Raised when URI scheme is invalid or not supported by the current backend.
```
pipermerriam: I think this docstring needs to be updated.  The *not supported* seems to now be covered by the `CannotHandleURI` exception.  Thoughts?

ethpm/package.py
```diff
@@ -98,39 +96,29 @@ def from_file(cls, file_path_or_obj: str, w3: Web3) -> "Package":
...
+        URI schemes supported:
+            - IPFS          `ipfs://Qm...`
+            - HTTP          `https://raw.githubusercontent.com/repo/path.json#hash`
+            - Registry      `ercXXX://registry.eth/greeter?version=1.0.0`
         """
-        ipfs_backend = get_ipfs_backend()
-        if ipfs_backend.can_handle_uri(ipfs_uri):
-            raw_package_data = ipfs_backend.fetch_uri_contents(ipfs_uri)
-            package_data = json.loads(to_text(raw_package_data))
+        good_backends = get_backends_for_uri(uri)
```
pipermerriam: Maybe just call this `backends` since there isn't another set of `bad_backends` in scope here.

ethpm/backends/registry.py
```diff
@@ -0,0 +1,39 @@
...
+        return is_valid_registry_uri(uri)
+
+    def fetch_uri_contents(self, uri: str, w3: Web3) -> bytes:
+        """
+        Return the contents stored at `uri` assuming `uri`
+        is a valid Registry URI pointing towards an IPFS URI.
+        Requires a valid web3 instance connected to the chain on which
+        the registry lives.
+        """
+        manifest_uri = lookup_manifest_uri_located_at_registry_uri(uri, w3)
```
pipermerriam: Is this the only place where the `w3` instance is actually used?  It'd be nice to get it out of the signature for `fetch_uri_contents` and have it be provided via some other mechanism.  Maybe as a constructor argument to this backend?

ethpm/utils/backend.py
```diff
@@ -0,0 +1,38 @@
...
+    default_ipfs = get_ipfs_backend_class()
+    good_backends = [
+        backend
+        for backend in URI_BACKENDS
+        # type ignored because of conflict with instantiating BaseURIBackend
+        if backend().can_handle_uri(uri)  # type: ignore
+    ]
+    if default_ipfs in good_backends:
+        # Move default to front of the list of good backends.
+        good_backends.insert(0, good_backends.pop(good_backends.index(default_ipfs)))
```
pipermerriam: Not sure if it's better, but here's how I would write this.

```python

def get_backend_for_uri(uri):
    backends = _get_backend_for_uri(uri)
    if not backends:
        raise CannotHandleURI(...)
    return backends

@to_tuple
def _get_backends_for_uri(uri):
    # special case the default IPFS backend to the first slot.
    default_ipfs = get_ipfs_backend_class()
    if default_ipfs in URI_BACKENDS and default_ipfs().can_handle_uri(uri):
        yield default_ipfs
    for backend_class in URI_BACKENDS:
        if backend_class is default_ipfs:
            continue
        elif backend_class().can_handle_uri(uri):
            yield backend_class
```

I *think* the logic of the function is at least *as* easy to understand in this format and it gets rid of the mutation.

pipermerriam:
Correct me if I'm wrong here:  I think the passing around of the `w3` instance is specifically for the on-chain registry backend.  If so, here's my proposal for how we can kick that can down the road a bit.

What do you think about defaulting the registry backend to use its own web3 instance that's by default connected to infura?

ethpm/exceptions.py
njgheorghita: Yeah true, there is a subtle difference between the two, but probably not worth the confusion between having two separate but similar exception classes, and the subtle difference is better explained through each instance's error message.

ethpm/backends/registry.py
pipermerriam: > Also, it's possible that RegistryURIBackend.fetch_uri_contents simply returns the ipfs uri stored on the registry and then we handle fetching the ipfs uri's contents outside of this backend -- but I like the idea of this method actually fetching and returning the stored ipfs uri's contents and encapsulating all of that logic here.

I think this concept exists within the populus code which dealt with packages.  There was a mechanism that allowed backends to *resolve* a package identifier, but resolution had a *finality* concept which indicated whether the resource it was returning was fully resolved, or whether it needed to be passed through another round of backend resolution.

pipermerriam: Here it is:

https://github.com/ethereum/populus/blob/feat/v2/populus/utils/packaging.py#L442

njgheorghita:
@pipermerriam I'm down to default to an infura web3 for the registry backend. What would that look like in terms of the api key? Just generate a new one strictly for this repo to use and if it gets abused/becomes problematic then come up with a fix?

pipermerriam:
RE infura and API keys.

Yes, I think that's an acceptable short term solution while this stuff is still experimental.

njgheorghita:
@pipermerriam Could you give this another :eyes: when you get the chance. Switched up the uri handling somewhat significantly to mirror the populus strategy more closely. 
`can_handle_uri` became ...
`can_resolve_uri` if uri points to a manifest (basically any ipfs uri) / `can_translate_uri` if uri points to another uri (basically any registry uri)

pipermerriam:
:crying_cat_face: sorry, got lost in a storm of github notifications. Looking now for anything that should get cleaned up post merge

ethpm/utils/backend.py
```diff
@@ -0,0 +1,73 @@
...
+    IPFSGatewayBackend,
+    InfuraIPFSBackend,
+    DummyIPFSBackend,
+    LocalIPFSBackend,
+    RegistryURIBackend,
+]
+
+
+def resolve_uri_contents(uri: str, fingerprint: bool = None) -> bytes:
+    is_resolvable = get_resolvable_backends_for_uri(uri)
```
pipermerriam: Just a naming convention thing, but I expected this to be a boolean and then was surprised when I saw it being iterated over.  Maybe something like `backends_for_resolution` or `resolver_backends/translator_backends`.

ethpm/utils/backend.py
```diff
@@ -0,0 +1,73 @@
...
+
+    raise CannotHandleURI(
+        "URI: {0} cannot be resolved by any of the available backends.".format(uri)
+    )
+
+
+@to_tuple
+def get_translatable_backends_for_uri(
+    uri: str
+) -> Generator[Type[BaseURIBackend], None, None]:
```
pipermerriam: Try replacing `Type[BaseURIBackend]` with:

```python
BackendType = typing.TypeVar('BackendType', bound=Type[BaseURIBackend)
```

I'm not sure if it will work, but the `bound` keyword IIRC is for specifying the lower bound on base classes that can be used.

pipermerriam: Also, cc @cburgdorf who has been our local type ninja.

ethpm/validation.py
```diff
@@ -46,12 +59,14 @@ def validate_registry_uri(uri: str) -> None:
...
-        validate_registry_uri_version(query)
-    validate_package_name(pkg_name[1:])
+    try:
+        validate_registry_uri_scheme(scheme)
+        validate_registry_uri_authority(authority)
+        if query:
+            validate_registry_uri_version(query)
+        validate_package_name(pkg_name[1:])
+    except ValidationError:
+        raise CannotHandleURI("{0} is not a valid registry URI.".format(uri))
```
pipermerriam: Hrm, this seems to violate my personal "principle of least surprise".  I expect validation functions to do exactly one of two things.

- return `None`
- raise a `ValidationError`

I'd expect the `CannotHandleURI` exception to occur at a different level in the code.

ethpm/utils/backend.py
njgheorghita: Haven't been able to get this working, even though after reading up on bound, it seems like this should work. 
with `BackendType = typing.TypeVar('BackendType', bound=Type[BaseURIBackend])`
I haven't been able to get around this error ... 
`(actual type "Type[BaseURIBackend]", expected type "BackendType")`
seems like `BackendType` isn't being evaluated properly? @cburgdorf any thoughts on what I'm doing wrong?

PR #80, opened 2018-07-24
Cache package `Deployment` and `Dependencies` properties

njgheorghita:
Branched from #76 

### What was wrong?
`Package.get_deployments` and `Package.get_build_dependencies` won't change throughout a `Package`'s life - so it makes little sense to call that method over and over

### How was it fixed?
Cached the `Deployments` obj and `Dependencies` obj as properties on the `Package` obj using the `cached_property` from Django
https://docs.djangoproject.com/en/2.0/_modules/django/utils/functional/#cached_property

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/utils/cache.py
+++ b/ethpm/utils/cache.py
@@ -1,0 +1,29 @@
+import functools
+from typing import Any, Callable
+
+
+class cached_property:
+    """
+    Decorator that converts a method with a single self argument into a
+    property cached on the instance.
+
+    Optional ``name`` argument allows you to make cached properties of other
+    methods. (e.g.  url = cached_property(get_absolute_url, name='url') )
+    """
+
+    def __init__(self, func: Callable[..., Any], name: str = None) -> None:
+        self.func = func
+        self.name = name or func.__name__
+        # type ignored b/c self is 'cached_property' but expects 'Callable[..., Any]'
+        functools.update_wrapper(self, func)  # type: ignore
+
+    def __get__(self, instance: Any, cls: Any = None) -> Any:
+        """
+        Call the function and put the return value in instance.__dict__ so that
+        subsequent attribute access on the instance returns the cached value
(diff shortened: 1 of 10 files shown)


ethpm/utils/cache.py
```diff
@@ -0,0 +1,27 @@
...
+    Decorator that converts a method with a single self argument into a
+    property cached on the instance.
+
+    Optional ``name`` argument allows you to make cached properties of other
+    methods. (e.g.  url = cached_property(get_absolute_url, name='url') )
+    """
+
+    def __init__(self, func: Callable[[Any], Any], name: str = None) -> None:
+        self.func = func
+        self.__doc__ = getattr(func, "__doc__")
```
pipermerriam: It might be better to use [`functools.update_wrapper`](https://docs.python.org/3.5/library/functools.html#functools.update_wrapper) to do this.  It'll move all of `__name__`, `__module__`, `__qualname__`, `__doc__` and `__annotactions__` as well as mergind `__dict__` on the two.  I think all of these are good things.

tests/conftest.py
```diff
@@ -109,7 +108,7 @@ def manifest_with_matching_deployment(w3, tmpdir, safe_math_manifest):
             "block": "0x420cb2b2bd634ef42f9082e1ee87a8d4aeeaf506ea5cdeddaa8ff7cbf911810c",
         }
     }
-    return manifest
+    return manifest, w3
```
pipermerriam: Why return `w3` also?  Seems like you could just use the `w3` fixture wherever you use this fixture to get the `w3` instance?  Or are there multiple `w3` instances floating around?

njgheorghita:
@pipermerriam Thanks for the review - it's actually branched off of pr #76 which is still waiting review - is there a better way to handle this? Probably leave off the `ready for review` label until the preceding pr is merged and I can rebase?

PR #82, opened 2018-08-02
Implement generate_file_hash ipfs util

njgheorghita:
### What was wrong?
Generating an ipfs hash for a file would be a useful util for testing

### How was it fixed?
stole IPFS util: `generate_file_hash` from https://github.com/ethereum/populus/blob/feat%2Fv2/populus/utils/ipfs.py

#### Cute Animal Picture
image


njgheorghita:
@pipermerriam ping when you get the chance - should be pretty quick as it was mostly lifted from populus

setup.py
```diff
@@ -61,6 +61,7 @@
         'eth-utils>=1.0.2,<2',
         'ipfsapi>=0.4.3,<1',
         'jsonschema>=2.6.0,<3',
+        "protobuf>=3.0.0",
```
pipermerriam: Can you put an upper bound on this so we don't accidentally do a breaking upgrade.

PR #83, opened 2018-08-07
Programmatically generate manifest

njgheorghita:
### What was wrong?
Feature: Programmatically generate manifest from solc output
Still more or less a WIP but could use some feedback on whether or not it seems headed in the right direction.

### How was it fixed?
Wrote class to manage manifest generation. 

#### Cute Animal Picture
image


njgheorghita:
@pipermerriam Ready for an initial review on this when you get the chance. Only super simple use cases are covered so far, any input on the best way to extend this to support handling options that come with bytecode objects, deployments, etc. would be great!

ethpm/manifest.py
```diff
@@ -0,0 +1,149 @@
...
+                )
+
+    @to_dict
+    def _create_contract_type(
+        self, contract_name: str, contract: str, solc: Dict[str, Any]
+    ) -> Generator[Tuple[str, Any], None, None]:
+        yield contract_name, self._generate_contract_type_dict(contract, solc)
+
+    @to_dict
+    def _generate_contract_type_dict(
```
pipermerriam: All of these that don't actually use `self` are ones I'd recommend moving to utility functions (just here in this module is fine).

ethpm/manifest.py
```diff
@@ -0,0 +1,149 @@
...
+        solc = load_json_from_file_path(self.solc_path)
+        for contract in solc["contracts"]:
+            contract_path, contract_name = contract.split(":")
+            source_data = self._create_source(contract_path)
+            setattr(self, "sources", {**self.sources, **source_data})
+            if contract_types is not None and contract_name in contract_types:
+                contract_type_data = self._create_contract_type(
+                    contract_name, contract, solc
+                )
+                setattr(
```
pipermerriam: Why not just `self.contract_types = {**self.contract_types, **contract_type_data}`.  Is the `setattr` there for a reason?

Also, this is equivalent to `self.contract_types = cytoolz.merge(self.contract_types, contract_type_data)`

It looks like you're incrementally building a dictionary here.  Maybe You could do that with a comprehension or another `@to_dict` helper function?

ethpm/manifest.py
```diff
@@ -0,0 +1,149 @@
...
+        Manifest is written to the same dir in which the `contracts/` dir is located.
+        """
+        manifest_contents = self.minified()
+        path = self.solc_path.parent / str(self.version + ".json")
+        path.write_text(manifest_contents)
+        return path
+
+    @to_dict
+    def _generate_attr_dict(self) -> Generator[Tuple[str, Any], Any, None]:
+        for key, value in self.__dict__.items():
```
pipermerriam: I think this is going to blow up on you.  I'd recommend just doing this explicitely.  There are a fixed number of top level keys and it should be reasonably sane to just list them all.

ethpm/manifest.py
```diff
@@ -0,0 +1,171 @@
...
+        yield "./{0}".format(contract_path), source_data
+
+
+def pin_single_source(
+    contract_path: str, solc_path: str, ipfs_backend: BaseIPFSBackend
+) -> str:
+    """
+    Return an IPFS URI, after pinning a source file to IPFS via `ipfs_backend`.
+    """
+    [ipfs_return_data] = ipfs_backend.pin_assets(Path(solc_path).parent / contract_path)
```
pipermerriam: This syntax confused me as I've not seen using list for unpacking return values.  Would it bother you to change this to tuple form?

```python
(x,) = ...
```

This is absolutely a nitpicky thing...

pipermerriam: In a non-nitpicky thread....

For an MVP, nothing here needs to change, however, I *think* that it would be good for us to think about the lifecycle for manifests.  For packaging, it is normally split into two distinct steps.  `build -> publish`.  This however results in partial publication as part of what I perceive to be the "build" step.  We should explore how we can do a full build, and push this publishing step out till the point where the publish the manifest itself.  Simply writing up an issue for this would be adequate for now so that we don't lose track of it.

ethpm/validation.py
```diff
@@ -33,6 +33,16 @@ def validate_package_name(pkg_name: str) -> None:
         raise ValidationError("{0} is not a valid package name.".format(pkg_name))
 
 
+def validate_manifest_version(version: str) -> None:
+    """
+    Raise an exception if the version is not 2.
+    """
+    if not int(version) == 2:
```
pipermerriam: I don't think it should be cast to an integer here as this would allow for a non-integer literal `2` to pass validation.

tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+from pathlib import Path
+
+PACKAGING_EXAMPLES_DIR = Path(__file__).parent / "ethpm" / "packaging-examples"
```
pipermerriam: Our test modules should not have `__init__` files (and thus they won't be proper modules.  You can put this in a `tests/conftest.py` and expose it as a fixture if you need access to the variable in tests.

tests/ethpm/packaging-examples/owned/1.0.0-pretty.json
```diff
@@ -0,0 +1,21 @@
+{
+  "manifest_version": "2",
+  "version": "1.0.0",
+  "package_name": "owned",
+  "meta": {
+    "license": "MIT",
+    "authors": [
+      "Piper Merriam <pipermerriam@gmail.com>"
```
pipermerriam: Side note: you guys should probably replace all of these with `author@example.com` or some other acceptable lorem-ipsum style address.

njgheorghita:
@pipermerriam Converted this to a more functional style of manifest generation as suggested. Followed the style used [to link bytecode](https://github.com/ethpm/py-ethpm/blob/e391af716f752d7a688dda64bb065a9bdb9eaad7/ethpm/contract.py#L96). When you get some time, any thoughts/suggestions for improvement would be super!

ethpm/manifest.py
```diff
@@ -0,0 +1,257 @@
...
+    for key, value in meta.items():
+        if key in META_FIELDS:
+            if type(value) is not META_FIELDS[key]:
+                raise ValidationError(
+                    "Values for {0} are expected to have the type {1}, instead got {2}.".format(
+                        key, META_FIELDS[key], type(value)
+                    )
+                )
+            continue
+        if allow_extra_meta_fields:
```
pipermerriam: This should be an `elif`.  The current code is functionally equivalent since the previous `if` block has a `continue` but a refactor could change that and cause breakage.

ethpm/manifest.py
```diff
@@ -0,0 +1,257 @@
...
+def build_sources(
+    compiled_json_output_path: Path, manifest: Dict[str, Any]
+) -> Dict[str, Any]:
+    """
+    Generate all source keys/values found in the linked compiled json output, and
+    attach them to the manifest.
+    """
+    if compiled_json_output_path:
+        source_data = generate_all_sources(compiled_json_output_path)
+        new_manifest = cytoolz.assoc(manifest, "sources", source_data)
```
pipermerriam: These might be more legible without the variable and just returning directly from here.

PR #86, opened 2018-08-10
Add IPFS node to circle environment

njgheorghita:
### What was wrong?
Better testing of IPFS integration was needed. 

### How was it fixed?
Configured circle to start a ci node in the background we can run some integration tests against

#### Cute Animal Picture
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,17 @@ common: &common
         name: install submodules
         command: git submodule update --init
     - run:
+        name: install ipfs
+        command:
+          wget https://dist.ipfs.io/go-ipfs/v0.4.11/go-ipfs_v0.4.11_linux-amd64.tar.gz &&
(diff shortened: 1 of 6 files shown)


njgheorghita:
@pipermerriam Not sure what's the best way forward here. Circle now has an ipfs node available to run integration tests against. But, the tests will fail if a user runs them locally and doesn't have their own IPFS node running. Thoughts?

pipermerriam:
I'd lump the tests into a separate test run and then add a pytest cli argument like this: https://github.com/ethereum/py-evm/blob/master/tests/trinity/conftest.py#L27-L29.  You can then conditionally skip the tests based on the presence of that flag.  Probably good to document this in the developer documentation.

njgheorghita:
@pipermerriam Thanks, made the change. I was getting some weird behavior with Circle for a bit where it wasn't recursing through all the test files after adding the flag to the pytest command, but seems to be working fine now.

PR #87, opened 2018-08-15
Refresh deployments when changing w3 attribute on a package

njgheorghita:
### What was wrong?
Deployments was a cached property on Package, but if you update the `w3` instance belonging to a `Package` it needs to update the `deployments` property to reflect the deployments available on the new chain.

### How was it fixed?
I'd like to test that switching between w3 instances on separate chains would return different deployments, but I'm not quite if that's possible, since it doesn't seem possible to have two web3 instances (instantiated with EthereumTesterProvider) that have different blockchain uris - to simulate switching between chains and their respective deployments.

#### Cute Animal Picture
image


voith:
For me the only thing that fixed the issue was installing from master. Only possible solution that I can think of right now is to create a new release in `eth-typing` and then retry with the new version.

pipermerriam:
@njgheorghita I accidentally broke everything for a few minutes.  Turns out that `eth-typing==1.2.0` isn't actually a thing and must have been an artifact from a failed release I made at some point.  When I *pushed* it to pypi this morning a bunch of things broke because it was actually an older version of the library.  I've since removed it from pypi and the github repo.

You *should* be able to do a fresh install of `eth-typing>=1.1.0`.  What specifically is the error your encountering?

ethpm/package.py
```diff
@@ -58,6 +61,8 @@ def set_default_w3(self, w3: Web3) -> None:
         """
         Set the default Web3 instance.
         """
+        if "deployments" in self.__dict__:
+            del self.deployments
```
pipermerriam: I have an alternate suggestion that I'm not sure is better.  **If** we stick with this one, this deserves a comment to explain why, and probably another comment in the actual `deployments` property function which explains that the cache gets busted via this mechanism.

My idea is to cache the deployments using a cache key based on the web3 instance itself.  That way, when a new web3 instance gets added, the cache is automatically busted.

```python
@property
def deployments(self):
    should_bust_cache = (
        self.deployments is None or
        self.deployments[0] != id(self.w3)
    )
    if should_bust_cache:
        deps = ...  # construct the deps
        self.deployments = (id(w3), deps)
    return self.deployments[1]
```

This code is pretty gross but hopefully it demonstrates the idea.

njgheorghita: Tinkered around with this idea, but running into an infinite recursion loop (`should_bust_cache` looks up the deployments property, calling the method and `should_bust_cache` again . . .) Know any trick to avoid this?

It would be nice to have a simpler cache strategy (so we can get rid of the `cached_property` util) but, I'm leaning towards leaving it as is - and adding some comments. 

Also occurred to me that we probably want to bust the `build_dependencies` cached property as well upon switching chains.

njgheorghita:
@pipermerriam The error is that `ContractName` isn't available to import (which was introduced in the `v1.1.0` release)

I pinned `eth-typing` specifically to `v1.1.0` here, but I'm still getting the same error (you can check it out in the broken circle builds here). It says that eth-typing `v.1.1.0` and eth-utils `v1.1.2` are installed, but my best guess is that pypi is still confused? I'm thinking @voith 's suggestion of cutting a new `eth-typing` release is the easiest way to sort this out

PR #89, opened 2018-08-17
Contract subclass to handle link references

njgheorghita:
### What was wrong?
Contract factory subclass was needed to manage Package's which contain link references between, and adjust the bytecode accordingly. 

### How was it fixed?
Wrote a `LinkableContract` subclass that inherits from `web3.contract.Contract`. Implements following methods.
- `has_linkable_bytecode`
- `link_bytecode`
- `is_bytecode_linked` property

#### Cute Animal Picture
image


ethpm/contract.py
```diff
@@ -0,0 +1,79 @@
...
+                    )
+                    self.bytecode = new_bytecode
+
+        if self.runtime_link_refs:
+            for ref in self.runtime_link_refs:
+                validate_empty_bytes(
+                    self.bytecode_runtime, ref["offsets"], ref["length"]
+                )
+                link_address = to_canonical_address(attr_dict[ref["name"]])
+                for offset in ref["offsets"]:
```
pipermerriam: We can remove this loop and the repeated mutation using the following pattern.

```python
@curry
def apply_link_reference(offset, value, bytecode):
    ...  # returns the new bytecode with `value` put into the location indicated by `offset`

# Create a bunch of functions that individually will apply a single link reference offset.
link_fns = (
    apply_link_reference(offset, attr_dict[ref['name']])
    for ref in link_refs
    for offset in ref['offsets']
)
# pipe the original bytecode through all of those functions, end result is fully linked.
linked_bytecode = cytoolz.pipe(original_bytecode, *link_fns)
```

ethpm/contract.py
```diff
@@ -0,0 +1,128 @@
...
+    contract factories with link references in their package's manifest.
+    """
+
+    deployment_link_refs = None
+    runtime_link_refs = None
+    is_bytecode_linked = False
+
+    def __init__(self, address: bytes = None, **kwargs: Dict[str, Any]) -> None:
+        super(LinkableContract, self).__init__(address=address, **kwargs)
+
```
njgheorghita: @carver I'm having some trouble with subclassing `LinkableContract` as both a factory and an instance.  I'd like to have the `LinkableContract` methods (ie `has_linkable_bytecode`/`link_bytecode`) available on the factory. 

Right now the way i'm doing this is passing in the factory as the `self` arg (i.e. `factory.has_linkable_bytecode(factory)`) but i'd much rather just call `factory.has_linkable_bytecode()` (which now raises a `missing 1 posarg self argument` error), any thoughts on how I could do this?

carver: Discussed in side channel. tl;dr - use `@combomethod`

njgheorghita:
@pipermerriam ready for an initial :eyes: whenever you get the chance

ethpm/contract.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    Returns the new bytecode with `value` put into the location indicated by `offset` and `length`.
+    """
+    try:
+        validate_empty_bytes(offset, length, bytecode)
+    except ValidationError:
+        raise BytecodeLinkingError("Link references cannot be applied to bytecode")
+
+    new_bytes = (
+        # Ignore linting error b/c black conflict
```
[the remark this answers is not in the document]
njgheorghita: Probably not, formatting rules for whitespace around `:` are different b/w black and flake8. since black is super opinionated, i'm just ignoring the flake8 complaint here. Unless it's better to update `tox.ini`?

ethpm/contract.py
```diff
@@ -0,0 +1,119 @@
...
+        bytecode[:offset]
+        + to_canonical_address(value)
```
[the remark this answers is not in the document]
njgheorghita: Is it a problem to simply require a canonical address in `attr_dict`? Or do we want to accept a wider range of addresses, and then normalize any that should be?

njgheorghita:
@pipermerriam Thanks for the review! I beefed up the tests and got rid of the class mutation, let me know if there's anything else

ethpm/contract.py
```diff
@@ -0,0 +1,122 @@
...
+        for ref in link_refs
+        for offset in ref["offsets"]
+    )
+    linked_bytecode = cytoolz.pipe(bytecode, *link_fns)
+    return linked_bytecode
+
+
+@cytoolz.curry
+def apply_link_reference(
+    offset: int, length: int, value: bytes, bytecode: bytes
```
pipermerriam: I think you should drop the `length` parameter and derive it from the `value`.

```python
length = len(value)
```

Otherwise, you could get weird results coming out if the actual length is different from the expected length

PR #92, opened 2018-08-27
Force canonicalized addresses everywhere

njgheorghita:
### What was wrong?
Use canonicalized addresses everywhere (excepting inside manifests)

### How was it fixed?
Force canonicalized addresses via validation function wherever an address is taken as an argument.

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/validation.py
+++ b/ethpm/validation.py
@@ -1,12 +1,26 @@
 import re
+from typing import Any
 from urllib import parse
 
-from eth_utils import is_checksum_address
+from eth_utils import is_address, is_canonical_address, is_checksum_address
 
 from ethpm.constants import PACKAGE_NAME_REGEX, REGISTRY_URI_SCHEME
 from ethpm.exceptions import ValidationError
 from ethpm.utils.ipfs import is_ipfs_uri
 from ethpm.utils.registry import is_ens_domain
+
+
+def validate_address(address: Any) -> None:
+    """
+    Raise a ValidationError if an address is not canonicalized.
+    """
+    if not is_address(address):
+        raise ValidationError("Expected an address, got: {0}".format(address))
+    if not is_canonical_address(address):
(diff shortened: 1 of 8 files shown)


ethpm/validation.py
```diff
@@ -1,14 +1,25 @@
...
-from eth_utils import is_checksum_address
+from eth_utils import is_canonical_address, is_checksum_address
 
 from ethpm.constants import PACKAGE_NAME_REGEX, REGISTRY_URI_SCHEME
 from ethpm.exceptions import ValidationError
 from ethpm.utils.ipfs import is_ipfs_uri
 from ethpm.utils.registry import is_ens_domain
 
 
+def validate_address(address: bytes) -> None:
```
pipermerriam: Normally validation functions should accept `Any` since you could in theory pass anything in.

ethpm/validation.py
```diff
@@ -1,14 +1,25 @@
...
+    """
+    Raise a ValidationError if an address is not canonicalized.
+    """
+    if not is_canonical_address(address):
```
pipermerriam: Maybe add an extra layer of logic here.

```python
if not is_canonical_address(value):
    if not is_address(value):
        raise ValidationError("Expected an address, got: ...")
    else:
        raise ValidationError("Malformed address, must be canonical, blah blah blah")
```

tests/ethpm/test_deployments.py
```diff
@@ -1,4 +1,4 @@
-from eth_utils import to_bytes
+from eth_utils import to_bytes, to_canonical_address
```
pipermerriam: I keep forgetting this repo uses `black` when I see import statements not ordered as expected.  At first it threw me off but now I'm starting to like it (not worrying about this stuff and not having to maintain the weird subjective rulesets in my head)

PR #93, opened 2018-08-30
Manifest builder

njgheorghita:
### What was wrong?
Builder to allow for detailed manifest generation. 

#### Cute Animal Picture
image


pipermerriam:
waiting for green tests or explicit request to review (unless told otherwise)

njgheorghita:
@pipermerriam green test ping

njgheorghita:
@pipermerriam ping-a-ling in case you missed ^

ethpm/tools/builder.py
```diff
@@ -0,0 +1,332 @@
...
+    elif package_root_dir and (package_root_dir / source_path_suffix).is_file():
+        source_data = (package_root_dir / source_path_suffix).read_text()
+    else:
+        raise ManifestBuildingError(
+            "Contract source cannot be resolved, please make sure that the working "
+            "directory is set to the correct directory or provide `pacakge_root_dir`."
+        )
+
+    if "sources" not in manifest:
+        manifest["sources"] = {}
```
pipermerriam: This mutates the input which is something I think we were going to try and avoid.  See below for how to get around this.

ethpm/tools/builder.py
```diff
@@ -0,0 +1,332 @@
...
+
+    return assoc(
+        manifest,
+        "sources",
+        assoc(manifest["sources"], "./{0}".format(source_path_suffix), source_data),
```
pipermerriam: I think you can clean this up using `assoc_in` since it will lazily create the inner dictionary for you.

```python
assoc_in(manifest, ["sources", "./{0}".format(source_path_suffix)], source_data)
```

ethpm/tools/builder.py
```diff
@@ -0,0 +1,332 @@
...
+    path to the source as defined in the compiler output.
+    """
+    return _inline_source(name, compiler_output, package_root_dir)
+
+
+@curry
+def _inline_source(
+    name: str,
+    compiler_output: Dict[str, Any],
+    package_root_dir: str,
```
pipermerriam: Should the type here be `Optional[str]`?

ethpm/tools/builder.py
```diff
@@ -0,0 +1,332 @@
...
+    name: str,
+    compiler_output: Dict[str, Any],
+    ipfs_backend: BaseIPFSBackend,
+    package_root_dir: str,
+    manifest: Manifest,
+) -> Manifest:
+    names_and_paths = get_names_and_paths(compiler_output)
+    source_path = names_and_paths[name]
+    if package_root_dir:
+        (ipfs_data,) = ipfs_backend.pin_assets(package_root_dir / source_path)
```
pipermerriam: This seems like something worth running pre-validation on to check if the file is present and to throw a friendly error here where you have context of what is happening rather than the backend throwing a potentially harder to understand exception when it can't find the file.

ethpm/tools/builder.py
```diff
@@ -0,0 +1,332 @@
...
+    else:
+        cwd = Path(os.getcwd())
+        (ipfs_data,) = ipfs_backend.pin_assets(cwd / source_path)
+
+    if "sources" not in manifest:
+        manifest["sources"] = {}
```
pipermerriam: Same suggestion as above about removing this mutation and using `assoc_in` to lazily populate this inner dictionary.

ethpm/tools/builder.py
```diff
@@ -0,0 +1,332 @@
...
+    except KeyError:
+        raise ManifestBuildingError(
+            "Contract name: {0} not found in the provided compiler output.".format(name)
+        )
+
+    if not kwargs:
+        return assoc_in(manifest, ["contract_types", name], all_type_data)
+
+    # Handle single kwarg if "alias"
+    if "alias" in kwargs and len(kwargs) is 1:
```
pipermerriam: Maybe remove this from `kwargs` and make `alias` an independent argument that defaults to `None`?

ethpm/tools/builder.py
```diff
@@ -0,0 +1,332 @@
...
+#
+
+
+def contract_type(name: str, compiler_output: Dict[str, Any], **kwargs) -> Manifest:
+    """
+    Returns a copy of manifest with added contract_data field as specified by kwargs.
+    If no kwargs are present, all available contract_data found in the compiler output
+    will be included.
+
+    To include specific contract_data fields, add kwarg set to True (i.e. `abi=True`)
```
pipermerriam: Maybe instead of using `kwargs` for this they should just be explicitly listed?  This is under the assumption that we can enumerate the possible desired keys.

tests/ethpm/tools/test_builder.py
```diff
@@ -0,0 +1,268 @@
...
+        manifest_version("2"),
+        version("1.0.0"),
+        return_package(w3),
+    )
+    assert isinstance(package, Package)
+    assert package.version == "1.0.0"
+
+
+def test_builder_with_manifest_validation():
+    with pytest.raises(ValidationError):
```
pipermerriam: Maybe add a `match` kwarg to this to ensure the validation error contains the expected text (which I assume would include the string `"_invalid_package_name"`)

njgheorghita: Ooh, neat, didn't know about this. Seems like a pretty useful thing I should be using more often to make sure the correct error is being raised?

PR #95, opened 2018-08-31
Remove eth-typing dependency

njgheorghita:
### What was wrong?
Whatever issue is going on with the `eth-typing` dependency, is no longer worth the effort to debug. This is a temporary solution, but since there are only a couple types being imported from `eth-typing`, I'm re-implementing the `typing` module here and adding the types.  

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/package.py
+++ b/ethpm/package.py
@@ -1,7 +1,6 @@
 import json
 from typing import Any, Dict
 
-from eth_typing import Address, ContractName
 from eth_utils import to_canonical_address, to_text
 from web3 import Web3
 from web3.eth import Contract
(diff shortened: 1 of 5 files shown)


njgheorghita:
going ahead and merging since it's a minor fix, will revisit the `eth-typing` dependency before releasing stable alpha

njgheorghita:
No clue what's going wrong, doesn't seem to be a problem with the circle cache since I can replicate this locally when installing into a fresh venv. Just going to stick with this temp fix and hope the bug somehow fixes itself magically when I come back to it

PR #97, opened 2018-09-02
Validate linked references

### LINKED ISSUE #110: On naming things
Something I've been thinking about since the library got into link reference land is whether we can come up with less confusing names for things.

We don't **have to** use the same names for things as the spec does....

I'm thinking about this in the context of `link_references` and `link_dependencies`.  I pretty much always have to think about which is which and even then, I'm regularly not 100% sure I'm right.  Maybe we could rename those to something like:

- `Contract.unlinked_references`
- `Contract.linked_references`

Probably some better names out there if any other wordsmiths want to take a stab at this.

njgheorghita:
requires https://github.com/ethereum/pytest-ethereum/pull/5 to be merged and released before passing CI

### What was wrong?
Validation of linked references was needed, along with validation of tx_receipt data if present in deployment data

renamed:
`runtime_link_refs` => `linked_references`
`deployment_link_refs` => `unlinked_references`

### How was it fixed?
Added validation that for deployments with link-dependencies, their on-chain bytecode references the same instances specified in their manifest. 

#### Cute Animal Picture
image

fixes #110


njgheorghita:
@pipermerriam Works for me, i'll plan on 930ish @ galvanize wed/thurs

pipermerriam:
Is this still valid, should I review?

njgheorghita:
@pipermerriam it's waiting on releasing `pytest-ethereum` to pypi - i'll ping you once that happens

njgheorghita:
@pipermerriam `pytest-ethereum` is only `python>= 3.6` - it also seems as though web3 is moving on from 3.5 - is it a fair time to deprecate 3.5 in `py-ethpm`?

njgheorghita:
oh yeah and ping - this is ready for review

njgheorghita:
and this pr's ready for a review whenever you get the chance

pipermerriam:
> is it a fair time to deprecate 3.5 in py-ethpm?

Yes

ethpm/contract.py
```diff
@@ -16,8 +16,8 @@ class LinkableContract(Contract):
     contract factories with link references in their package's manifest.
     """
 
-    deployment_link_refs = []  # type: List[Dict[str, Any]]
-    runtime_link_refs = []  # type: List[Dict[str, Any]]
+    unlinked_references = []  # type: List[Dict[str, Any]]
```
pipermerriam: This will leak across instances of the class.  You should never use mutable data structures as class properties.  Since these are class level properties, these should be changed to be `tuple`, or just drop the default value and set them to `None` to represent empty.

ethpm/package.py
```diff
@@ -215,5 +222,50 @@ def deployments(self) -> Union["Deployments", Dict[None, None]]:
             )
             for deployment_data in deployments.values()
         }
+        validate_deployments_tx_receipt(deployments, self.w3)
+        linked_deployments = get_linked_deployments(deployments)
+        if linked_deployments:
+            for deployment_data in linked_deployments.values():
+                on_chain_bytecode = self.w3.eth.getCode(
+                    to_checksum_address(deployment_data["address"])
```
pipermerriam: Assuming the version of web3 you're using supports it, I think it'd be better to use `to_canonical_address` since this is technically programatic access.  Neither is really better, but semantically, I think using `bytes` addresses here is slightly more appropriate.

ethpm/package.py
```diff
@@ -215,5 +222,50 @@ def deployments(self) -> Union["Deployments", Dict[None, None]]:
...
+            raise BytecodeLinkingError(
+                "Expected build dependency: {0} not found in package's build dependencies.".format(
+                    value.split(":")[0]
+                )
+            )
+        # Find and return resolved, nested ref
+        else:
+            unresolved_linked_ref = value.split(":", 1)[-1]
+            build_dependency = self.build_dependencies[value.split(":")[0]]
+            return build_dependency._resolve_link_dependencies(unresolved_linked_ref)
```
pipermerriam: Confusing python thing.  `return` statements in generators don't return anything.

```python
In [7]: def a_generator(v):
   ...:     if v:
   ...:         yield "v is true"
   ...:     else:
   ...:         return "v is false"
   ...:

In [9]: tuple(a_generator(True))
Out[9]: ('v is true',)

In [10]: tuple(a_generator(False))
Out[10]: ()
```

These either all need to be `yield` statements or `return` statements and it looks like they could all be changed to be `return` statements.

ethpm/package.py
```diff
@@ -215,5 +222,50 @@ def deployments(self) -> Union["Deployments", Dict[None, None]]:
...
+                    to_checksum_address(deployment_data["address"])
+                )
+                unresolved_linked_refs = normalize_linked_references(
+                    deployment_data["runtime_bytecode"]["link_dependencies"]
+                )
+                resolved_linked_refs = tuple(
+                    self._resolve_linked_references(link_ref, deployments)
+                    for link_ref in unresolved_linked_refs
+                )
+                validate_linked_references(resolved_linked_refs[0], on_chain_bytecode)
```
pipermerriam: I'm confused about this `[0]` index.  Shouldn't validation happen for everything in that iterable?

ethpm/utils/deployments.py
```diff
@@ -0,0 +1,97 @@
...
+
+def get_linked_deployments(deployments: Dict[str, Any]) -> Dict[str, Any]:
+    """
+    Returns all deployments found in a chain URI's deployment data that contain link dependencies.
+    """
+    linked_deployments = {
+        dep: data
+        for dep, data in deployments.items()
+        if "runtime_bytecode" in data
+        and "link_dependencies" in data["runtime_bytecode"]
```
pipermerriam: just noting that you can also do this with `cytoolz.get_in(('runtime_bytecode', 'link_dependencies'], data)` but I'm not sure if it's actually a better approach.

ethpm/utils/deployments.py
```diff
@@ -0,0 +1,97 @@
...
+    }
+    for deployment, data in linked_deployments.items():
+        if deployment == data["runtime_bytecode"]["link_dependencies"][0]["value"]:
```
pipermerriam: The `[0]` indexing here confuses me as well.  Why do we have a sequence style data structure which only has one element?

ethpm/utils/deployments.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    Validate that address and block hash found in deployment data match what is found on-chain.
+    """
+    for name, data in deployments.items():
+        if "transaction" in data:
+            tx_hash = data["transaction"]
+            tx_receipt = w3.eth.getTransactionReceipt(tx_hash)
+            tx_address = tx_receipt["contractAddress"]
+
+            if to_canonical_address(tx_address) != to_canonical_address(
```
pipermerriam: `if not eth_utils.is_same_address(tx_address, data["address"])`

PR #99, opened 2018-09-08
Generate registry manifest

njgheorghita:
### What was wrong?

- Change builder to use compiler_output generated from standard-json solc output, rather than combined-json output to allow for better handling of link-references
- Expand builder to support more complicated use cases with link references in the contract-type // such as a registry manifest 
- removed `manifest.py` since that was an approach to building manifests that is made redundant by the builder

#### Cute Animal Picture

image


ethpm/tools/builder.py
```diff
@@ -328,9 +321,86 @@ def normalize_contract_type(
...
+
+
+@to_dict
+def normalize_link_ref(
+    link_ref: Dict[str, Any], bytecode: bytes
+) -> Generator[Dict[str, Any], None, None]:
+    for name, real_data in link_ref.items():
+        yield "name", name
+        yield "length", 20
+        yield "offsets", normalize_offsets(link_ref, bytecode)
```
pipermerriam: This whole function is confusing.  The `to_dict` implies that it is making a *single* dictionary.  The type hinting implies it is supposed to make many dictionaries.  I think this body inside the `for` loop should be yielding a single dictionary rather than the key/value format.  If you run this function I think you'll find it will only ever produce a single link reference rather than the iterable of them you are probably expecting.

ethpm/tools/builder.py
```diff
@@ -328,9 +321,86 @@ def normalize_contract_type(
...
+
+
+@to_list
+def normalize_offsets(
+    data: Dict[str, Any], bytecode: bytes
+) -> Generator[List[int], None, None]:
+    for link_ref in data.values():
+        for ref in link_ref:
+            validate_link_ref(ref["start"] * 2, ref["length"] * 2, bytecode)
```
pipermerriam: Hrm, baking this validation into here feels odd.  I'm assuming this is validating that the sections in the bytecode are indeed all `0`s?  If so, I'd remove this from here and keep this as a post-processing step once you've gathered all of the references.

njgheorghita: This is actually using the validate function from line 389 that validates the slot represented by the offset and length begins and ends with a `__` - or is a valid placeholder from the solidity compiler. I was thinking more of a pre-processing validation to make sure that the offsets are valid before replacing the placeholder with `0`s

ethpm/tools/builder.py
```diff
@@ -298,39 +291,109 @@ def _contract_type(
...
+        "length": 20,
+        "offsets": normalize_offsets(link_ref, bytecode),
+    }
+
+
+@to_list
+def normalize_offsets(data: Dict[str, Any], bytecode: str) -> Iterable[List[int]]:
+    for link_ref in data.values():
+        for ref in link_ref:
+            validate_link_ref(ref["start"] * 2, ref["length"] * 2, bytecode)
```
pipermerriam: I still think it's an anti-pattern to mix validation and normalization.  There's nothing inherently bad about this but I *think* it would be cleaner to keep these separate.

1. generate the normalized link_ref objects.
2. validate the offsets (all at once) against the original bytecode.

njgheorghita: Gotcha, moved the link_ref validation to it's own process on lines 347-351

tests/ethpm/tools/test_builder.py
```diff
@@ -263,6 +288,40 @@ def test_builder_with_standard_token_manifest(
...
+        manifest_version("2"),
+        version("1.0.0"),
+        pin_source("Authority", compiler_output, ipfs_backend),
+        pin_source("IndexedOrderedSetLib", compiler_output, ipfs_backend),
+        pin_source("PackageDB", compiler_output, ipfs_backend),
+        pin_source("PackageIndex", compiler_output, ipfs_backend),
+        pin_source("PackageIndexInterface", compiler_output, ipfs_backend),
+        pin_source("ReleaseDB", compiler_output, ipfs_backend),
+        pin_source("ReleaseValidator", compiler_output, ipfs_backend),
+        pin_source("SemVersionLib", compiler_output, ipfs_backend),
```
pipermerriam: Not for this PR but this looks like you should be able to do it as:

```python
pin_sources(
    "Authority", "PackageDB", "ReleaseDB", 
    compiler_output=compiler_output, backend=ipfs_backend,
)
# or maybe
pinner = pin_sources(compiler_output=compiler_output, backend=ipfs_backend)
build(
    pinner("Authority"),
    pinner("PackageDB"),
    ...
)
```

tests/ethpm/tools/test_builder.py
```diff
@@ -263,6 +288,40 @@ def test_builder_with_standard_token_manifest(
...
+        contract_type("Authorized", compiler_output, abi=True),
```
pipermerriam: Same thought here for reducing boilerplate.

```python
contract_types(
    "Authorized", "PackageDB",
    compiler_output=compiler_output, abi=True, ...
)
# or
extractor = contract_types(compiler_output)
build(
    extractor("PackageDB", abi=True, deployment_bytecode=True)
```

Just some thoughts on how this API could shape up.

njgheorghita:
@pipermerriam Yeah I think that's a cool idea, and something ppl would find useful.

njgheorghita:
@pipermerriam For the `check_manifest` function, were you thinking of something that is a part of the `build()` api? Or a separate helper function that you can pass any manifest through anytime?

PR #100, opened 2018-09-09
Write builder to_disk function

njgheorghita:
### What was wrong?
Builder function needed to write manifests to disk

### How was it fixed?
Wrote `to_disk(manifest_root_dir, manifest_name, prettify)`. 

#### Cute Animal Picture
image

### DIFF
--- a/docs/builder.rst
+++ b/docs/builder.rst
@@ -10,10 +10,15 @@ Cookbook
 To create a simple manifest
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-For all manifests, the following fields are *required*.
-    - ``package_name``      (i.e. "owned")
-    - ``version``           (i.e. "1.0.0")
-    - ``manifest_version``  (i.e. "2")
+For all manifests, the following ingredients are *required*.
+
+.. code:: python
+
+   {}
+   package_name(str)
+   version(str)
+   manifest_version(str)
+
 
 The builder (i.e. ``build()``) expects a dict as the first argument. This dict can be empty, or populated if you want to extend an existing manifest.
 
@@ -45,6 +50,10 @@ The builder (i.e. ``build()``) expects a dict as the first argument. This dict c
 To return a ``Package``
 ~~~~~~~~~~~~~~~~~~~~~~~
 
+.. code:: python
+
+   return_package(w3: Web3)
+
 By default, the manifest builder returns a dict representing the manifest. To return a ``Package`` instance (instantiated with the generated manifest) from the builder, add the ``return_package()`` builder function with a valid ``web3`` instance to the end of the builder.
 
 .. doctest::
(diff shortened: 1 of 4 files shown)


docs/builder.rst
```diff
@@ -10,10 +10,15 @@ Cookbook
...
-For all manifests, the following fields are *required*.
-    - ``package_name``      (i.e. "owned")
-    - ``version``           (i.e. "1.0.0")
-    - ``manifest_version``  (i.e. "2")
+For all manifests, the following ingredients are *required*.
+
+.. code:: python
+
+   {}
+   package_name(str)
```
pipermerriam: Maybe since these are all **always** required we should provide a helper for doing them in a single step.

```python
init_manifest(name, version, manifest_version):
    ...
```

Where `manifest_version` isn't required and defaults to the latest version?

docs/builder.rst
```diff
@@ -63,9 +72,12 @@ By default, the manifest builder returns a dict representing the manifest. To re
...
 
 
+To validate a manifest
+~~~~~~~~~~~~~~~~~~~~~~
+
+.. code:: python
 
-To validate the built manifest
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+   validate()
```
pipermerriam: These should probably be contextualized within a `built(...)` for these code examples to be the most helpful.

docs/builder.rst
```diff
@@ -92,16 +104,57 @@ By default, the manifest builder does *not* perform any validation that the gene
...
+       manifest_root_dir: Optional[Path],
+       manifest_name: Optional[str],
+       prettify: Optional[bool],
+   )
+
+
+Writes the active manifest to disk. Will not overwrite an existing manifest with the same name and root directory.
+
+Defaults
+- Writes manifest to cwd unless a ``Path`` is provided as manifest_root_dir.
```
pipermerriam: change `cwd` to ``os.getcwd`` or expand to be full "current working directory (as returned by ``os.getcwd()``)"

docs/builder.rst
```diff
@@ -92,16 +104,57 @@ By default, the manifest builder does *not* perform any validation that the gene
...
+   ...     {},
+   ...     package_name("owned"),
+   ...     manifest_version("2"),
+   ...     version("1.0.0"),
+   ...     to_disk(manifest_root_dir=p, manifest_name="manifest.json", prettify=True),
+   ... )
+   {'package_name': 'owned', 'manifest_version': '2', 'version': '1.0.0'}
+   >>> with open(str(p / "manifest.json")) as f:
+   ...     actual_manifest = f.read()
+   >>> assert actual_manifest == expected_manifest
```
pipermerriam: For a code example, maybe rather than populating `expected_manifest` you can just load the manifest and then do `print(manifest)`.  The doctest will do the correct comparison with the embedded terminal output.

docs/builder.rst
njgheorghita: sounds good, will add in a following pr

docs/builder.rst
njgheorghita: Yup, it's a couple lines below - I thought that just providing the "ingredient" for how to do something at the beginning of each section might be helpful for fast lookup rather than digging through the example below

pipermerriam: I was actually suggesting something as simple as:

```python
build(
    ...,
    validate(),
)
```

Just to show it is intended to be used within the `build` API.

njgheorghita: Ahh, gotcha cool will update in following pr

PR #101, opened 2018-09-11
Extend builder and write initial checker

njgheorghita:
### What was wrong?
- added `pin_sources()` and `inline_sources` fns to builder, to allow for creating a pinner/inliner from which you can pin/inline multiple sources
- added `init_manifest()` to builder for simpler init-ing
- started writing the `checker` tool

#### Cute Animal Picture
image

### DIFF
--- a/docs/builder.rst
+++ b/docs/builder.rst
@@ -14,10 +14,18 @@ For all manifests, the following ingredients are *required*.
 
 .. code:: python
 
-   {}
-   package_name(str)
-   version(str)
-   manifest_version(str)
+   build(
+       {},
+       package_name(str),
+       version(str),
+       manifest_version(str),
+       ...,
+   )
+   # Or
+   build(
+       init_manifest(package_name: str, version: str, manifest_version: str="2")
+       ...,
+   )
 
 
 The builder (i.e. ``build()``) expects a dict as the first argument. This dict can be empty, or populated if you want to extend an existing manifest.
@@ -46,15 +54,28 @@ The builder (i.e. ``build()``) expects a dict as the first argument. This dict c
    >>> assert built_manifest == expected_manifest
    >>> assert extended_manifest == expected_manifest
 
+With ``init_manifest()``, which populates "version" with "2" (the only supported EthPM specification version), unless provided with an alternative "version".
+
+.. doctest::
+
+   >>> build(
+   ...     init_manifest("owned", "1.0.0"),
+   ... )
+   {'package_name': 'owned', 'version': '1.0.0', 'manifest_version': '2'}
+
+
 
 To return a ``Package``
 ~~~~~~~~~~~~~~~~~~~~~~~
 
 .. code:: python
 
-   return_package(w3: Web3)
+   build(
+       ...,
+       as_package(w3: Web3),
+   )
 
-By default, the manifest builder returns a dict representing the manifest. To return a ``Package`` instance (instantiated with the generated manifest) from the builder, add the ``return_package()`` builder function with a valid ``web3`` instance to the end of the builder.
+By default, the manifest builder returns a dict representing the manifest. To return a ``Package`` instance (instantiated with the generated manifest) from the builder, add the ``as_package()`` builder function with a valid ``web3`` instance to the end of the builder.
 
 .. doctest::
 
@@ -67,7 +88,7 @@ By default, the manifest builder returns a dict representing the manifest. To re
    ...     package_name("owned"),
    ...     manifest_version("2"),
    ...     version("1.0.0"),
-   ...     return_package(w3),
+   ...     as_package(w3),
    ... )
    >>> assert isinstance(built_package, Package)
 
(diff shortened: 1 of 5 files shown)


njgheorghita:
@pipermerriam started writing the checker, for a manifest. How specific should these checks be? Now that i'm typing out loud, how about as specific as possible, but ignoring any type-checking? The assumption being that the checker validates the natural language spec and would be used in tandem with `validate()` (which checks against the json-spec).

Any thoughts on the general direction/api of the checker would be great

Also, i'm not sure we can follow the pinner/inliner pattern for contract sources. Since there are so many optional args, our only options that I could tell would be to either provide each `contract_type()` individually with it's associated compiler_output - or you can curry together an "extractor", but in that case the user has to explicitly provide all the bool args. i.e 

> extractor("Owned", abi=False, natspec=True, deployment_bytecode=False, runtime_bytecode=False, compiler=True ...)

docs/builder.rst
```diff
@@ -14,10 +14,19 @@ For all manifests, the following ingredients are *required*.
...
+       {},
+       package_name(str),
+       version(str),
+       manifest_version(str),
+       ...,
+   )
+
+   # Or
+   build(
+       init_manifest(package_name: str, version: str, manifest_version: str=None)
```
pipermerriam: The `None` should be replaced with `"2"` right?

docs/builder.rst
```diff
@@ -46,13 +55,26 @@ The builder (i.e. ``build()``) expects a dict as the first argument. This dict c
...
    >>> assert extended_manifest == expected_manifest
 
+With ``init_manifest()``, which populates "version" with "2" (the only supported EthPM specification version), unless provided with an alternative "version".
+
+.. doctest::
+
+   >>> initialized_manifest = build(
+   ...     init_manifest("owned", "1.0.0"),
+   ... )
+   >>> assert initialized_manifest == expected_manifest
```
pipermerriam: As code examples go, I think this one would be better if you dropped the `expected_manifest` and variable assignment and allowed the doctest to just test against the return value of `build(...)` as it dumps to the terminal.

```
.. doctest
    >>> build(init_manifest('owned', '1.0.0'))
    {...<the-raw-manifest-dict>...}
```

docs/builder.rst
```diff
@@ -46,13 +55,26 @@ The builder (i.e. ``build()``) expects a dict as the first argument. This dict c
...
 
 To return a ``Package``
 ~~~~~~~~~~~~~~~~~~~~~~~
 
 .. code:: python
 
-   return_package(w3: Web3)
+   build(
+       ...,
+       return_package(w3: Web3),
```
pipermerriam: maybe rename to `as_package`?  shorter and doesn't have the `return` keyword in it which can be visually confusing.

docs/builder.rst
```diff
@@ -290,11 +341,20 @@ To include the source as a content-addressed URI, ``Py-EthPM`` can pin your sour
...
    ...   }
    ... }
+   >>> # With `pin_source()`
    >>> built_manifest = build(
    ...     BASE_MANIFEST,
    ...     pin_source("Owned", compiler_output, ipfs_backend, package_root_dir=owned_dir),
    ... )
    >>> assert expected_manifest == built_manifest
+   >>> # With `pin_sources()` for multiple sources from the same compiler output
+   >>> pinner = pin_sources(compiler_output, ipfs_backend, package_root_dir=owned_dir)
```
pipermerriam: Maybe to avoid visual confusion between `pin_source` and `pin_sources` since they only differ by the trailing `s` we could rename `pin_sources -> source_pinner` which aligns better with the naming convention we are setting up here in the docs.

docs/builder.rst
```diff
@@ -269,11 +311,20 @@ To inline the source code directly in the manifest, use the ``inline_source()``
...
    ...   }
    ... }
+   >>> # With `inline_source()`
    >>> built_manifest = build(
    ...     BASE_MANIFEST,
    ...     inline_source("Owned", compiler_output, package_root_dir=owned_dir),
    ... )
    >>> assert expected_manifest == built_manifest
+   >>> # With `inline_sources()` for multiple sources from the same compiler output
+   >>> inliner = inline_sources(compiler_output, package_root_dir=owned_dir)
```
pipermerriam: same here with rename to `source_inliner`

ethpm/tools/checker.py
```diff
@@ -0,0 +1,262 @@
...
+        check_keywords(manifest["meta"]),
+        check_links(manifest["meta"]),
+    )
+    return build(warnings, *meta_validation)
+
+
+@cytoolz.curry
+def check_authors(meta, warnings):
+    if "authors" not in meta:
+        return assoc(warnings, "authors", WARNINGS["authors_missing"])
```
pipermerriam: Probably worth namespacing the string as `"meta.authors"` so that the `warnings` dictionary keys everything by the dot-separated-path that it would be looked up from the manifest.

njgheorghita:
Yeah, i'm not the biggest fan of those tests, and their coverage isn't as great as I'd like. But, like you said, I was hardpressed to come up with a better way that would be worth the time to implement. When I get around to extending the checker, i'll give it another go

PR #104, opened 2018-09-18
Add `needs_bytecode_linking` to `LinkableContract`

njgheorghita:
### What was wrong?
Handling link references strategy was insufficient for pre-linked contract factories. 

This is needed for the `linker` strategy introduced to `pytest-ethereum` [here]().

### How was it fixed?
Replaced `Contract.has_linkable_bytecode` and `Contract.is_bytecode_linked` with `Contract.needs_bytecode_linking` that will automatically detect if a contract's link references are already linked.

some thoughts on possible improvements...
- option to validate link-refs against an address rather than simply asserting they are not `0`s 
- support partially pre-linked contract factories (right now every link-ref has to be unlinked, otherwise we assume that the factory is entirely linked.

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/contract.py
+++ b/ethpm/contract.py
@@ -1,7 +1,9 @@
 from typing import Any, Dict, List, Type  # noqa: F401
 
 import cytoolz
-from eth_utils import is_canonical_address, to_checksum_address
+from eth_utils import is_canonical_address, to_bytes, to_checksum_address
+from eth_utils.toolz import assoc
+from web3 import Web3
 from web3.contract import Contract
 from web3.utils.decorators import combomethod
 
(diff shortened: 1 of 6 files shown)


ethpm/contract.py
```diff
@@ -17,10 +17,9 @@ class LinkableContract(Contract):
 
     deployment_link_refs = []  # type: List[Dict[str, Any]]
     runtime_link_refs = []  # type: List[Dict[str, Any]]
-    is_bytecode_linked = False
 
     def __init__(self, address: bytes = None, **kwargs: Dict[str, Any]) -> None:
-        if self.has_linkable_bytecode() and not self.is_bytecode_linked:
+        if self.needs_bytecode_linking():
```
pipermerriam: I've found that non-property based boolean checks like this are often error prone in python because `if self.needs_bytecode_linking` will always return truthy (since `function` objects are considered *truthy* in python.  Maybe this should be `is_bytecode_fully_linked`?

ethpm/contract.py
```diff
@@ -31,45 +30,35 @@ def __init__(self, address: bytes = None, **kwargs: Dict[str, Any]) -> None:
...
-        Return a boolean indicating the presence of linkable bytecode
-        on this contract factory or instance.
-        """
-        if self.deployment_link_refs or self.runtime_link_refs:
-            return True
-        return False
-
-    @combomethod
-    def linked_contract_types(self) -> List[str]:
+    @classmethod
```
pipermerriam: Ah, I see now that this has to be a class method.  How about computing this at the time of class creation?  This *could* be done either within the `Contract.factory` method or as a metaclass (but probably easier if done in the `Contract.factory`)

ethpm/contract.py
```diff
@@ -105,6 +93,20 @@ def validate_attr_dict(self, attr_dict: Dict[str, str]) -> None:
...
 
+def is_prelinked_bytecode(bytecode: bytes, link_refs: List[Dict[str, Any]]) -> bool:
+    """
+    Returns False if all expected link_refs are unlinked, otherwise returns True.
+    todo support partially pre-linked bytecode (currently all or nothing)
+    """
+    for link_ref in link_refs:
+        for offset in link_ref["offsets"]:
+            try:
+                validate_empty_bytes(offset, link_ref["length"], bytecode)
```
pipermerriam: This is *probably* an **ok** temporary solution, but I don't think that checking for all `0`'s in the link locations should be the final solution since it is valid to link against the zero address.  The *final* solution should probably be cross checking between `link_dependencies` and `link_references` and seeing which elements from the `link_references` are not satisfied in the `link_dependencies`  (it's possible I've reversed those two, I get confused as to which one represents the "it was linked with these values" and which one is "these are the places that need linking".)

ethpm/package.py
```diff
@@ -190,12 +190,13 @@ def build_dependencies(self) -> "Dependencies":
     #
 
     @cached_property
-    def deployments(self) -> "Deployments":
+    def deployments(self) -> Union["Deployments", Dict[None, None]]:
```
pipermerriam: The `Dict[None, None]` is confusing... I think it's intended to represent returning an empty dictionary.  Maybe it'd be better to have the return value be `None` in that case, though that does mean you have to do return value checking which isn't ideal.

njgheorghita: Yup, leaning towards just leaving the confusing type in for now to make life slightly easier in `pytest-ethereum` re. return value checking

ethpm/contract.py
njgheorghita: Agreed, put it on the list - will work on improving this in a future pr

ethpm/contract.py
```diff
@@ -30,46 +30,33 @@ def __init__(self, address: bytes = None, **kwargs: Dict[str, Any]) -> None:
...
 
     @classmethod
-    def constructor(self, *args: Any, **kwargs: Any) -> bool:
-        if self.has_linkable_bytecode() and not self.is_bytecode_linked:
+    def factory(cls, web3, class_name=None, **kwargs):
+        cls.needs_bytecode_linking = False
+        if "deployment_link_refs" in kwargs and kwargs['deployment_link_refs']:
+            if not is_prelinked_bytecode(to_bytes(hexstr=kwargs['bytecode']), kwargs['deployment_link_refs']):
+                cls.needs_bytecode_linking = True
+        return super(LinkableContract, cls).factory(web3, class_name, **kwargs)
```
njgheorghita: @pipermerriam can I get another :eyes: on this. Not sure if this is the best way to implement the `needs_bytecode_linking` factory. It's working as intended, but especially w/ line 69-70 feels kind of hacky to me

pipermerriam: Should look something like this.

```python
class LinkableContract(Contract):
    ...
    # this is what allows this value to be set with `kwargs` in the `factory` method
    needs_bytecode_linking = None

    @classmethod
    def factory(cls, web3, class_name=None, **kwargs):
        needs_bytecode_linking = ...  # whatever check needs to be done to figure this out
        kwargs = assoc(kwargs, 'needs_bytecode_linking', needs_bytecode_linking)
        return super(LinkableContract, cls).factory(web3, class_name, **kwargs)
```

The main `factory` method uses `PropertyCheckingFactory` which is a metaclass which injects all of the given `kwargs` as class properties, but **only if** the class or one of the base classes already has the property (to restrict accidental injection of new class properties due to typos, etc).

ethpm/contract.py
```diff
@@ -79,7 +66,8 @@ def link_bytecode(cls, attr_dict: Dict[str, str]) -> Type["LinkableContract"]:
         linked_class.bytecode_runtime = apply_all_link_references(
             linked_class.bytecode_runtime, linked_class.runtime_link_refs, attr_dict
         )
-        linked_class.is_bytecode_linked = True
+        linked_class.needs_bytecode_linking = False
+        cls.needs_bytecode_linking = True
```
pipermerriam: This doesn't seem right.

Instead of building `linked_class` and then mutating it with the newly linked bytecodes, can we instead link the values that need linking and gather up the newly linked `bytecode` and ` runtime_bytecode` values.  Then pass them into `factory`?

```python
runtime = apply_all_link_references(...)
bytecode = apply_all_link_references(...)
linked_class = cls.factory(cls.web3, bytecode_runtime=runtime, bytecode=bytecode)
return linked_class
```

We can add a sanity check that the bytecode is indeed linked before exiting by checking `linked_class.needs_bytecode_linking` which should be set to `False` if everything went well.

PR #108, opened 2018-09-24
Write to_ipfs function for builder

njgheorghita:
### What was wrong?
It seemed like a `to_ipfs` function in the builder - that would automatically pin a manifest - would be useful

### How was it fixed?
wrote `to_ipfs` builder function
temporarily deprecated `IPFSGatewayBackend` since it doesn't expose a r&w api

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/tools/builder.py
+++ b/ethpm/tools/builder.py
@@ -1,10 +1,10 @@
 import functools
-import json
 from pathlib import Path
+import tempfile
 from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Tuple
 
 import cytoolz
-from eth_utils import add_0x_prefix, to_dict, to_list
+from eth_utils import add_0x_prefix, to_bytes, to_dict, to_list
 from eth_utils.toolz import assoc, assoc_in, curry, pipe
 from web3 import Web3
 
@@ -12,7 +12,10 @@ from ethpm import Package
 from ethpm.backends.ipfs import BaseIPFSBackend
 from ethpm.exceptions import ManifestBuildingError, ValidationError
 from ethpm.typing import Manifest
-from ethpm.utils.manifest_validation import validate_manifest_against_schema
+from ethpm.utils.manifest_validation import (
+    format_manifest,
+    validate_manifest_against_schema,
+)
 
 
 def build(obj: Dict[str, Any], *fns: Callable[..., Any]) -> Dict[str, Any]:
@@ -476,7 +479,7 @@ def as_package(w3: Web3, manifest: Manifest) -> Package:
     return Package(manifest, w3)
 
 
-def to_disk(
+def write_to_disk(
     manifest_root_dir: Optional[Path] = None,
     manifest_name: Optional[str] = None,
     prettify: Optional[bool] = False,
@@ -489,11 +492,11 @@ def to_disk(
       manifest name (which must end in json) is provided as manifest_name.
     - Writes the minified manifest version to disk unless prettify is set to True.
     """
-    return _to_disk(manifest_root_dir, manifest_name, prettify)
+    return _write_to_disk(manifest_root_dir, manifest_name, prettify)
 
 
 @curry
-def _to_disk(
+def _write_to_disk(
     manifest_root_dir: Optional[Path],
     manifest_name: Optional[str],
     prettify: Optional[bool],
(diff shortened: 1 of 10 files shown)


ethpm/tools/builder.py
```diff
@@ -524,14 +527,35 @@ def _to_disk(
...
+    Returns the IPFS pin data after pinning the manifest to the provided IPFS Backend.
+
+    `to_ipfs()` Should *always* be the last argument in a builder, as it will return the pin
+    data and not the manifest.
+    """
+    return _to_ipfs(ipfs_backend, prettify)
+
+
+@cytoolz.curry
+def _to_ipfs(
```
pipermerriam: I don't think you need the two functions.  That pattern is only really necessary when you need to do `*args`.  I think you can just reverse the argument order here so that `manifest` is the first argument and then require the others to be keyword

```python
def to_ipfs(manifest, *, ipfs_backend, prettify=False):
    ...
```

This will have the elegant effect of requiring both `prettify` and `ipfs_backend` to be passed as keywoard arguments.

ethpm/tools/builder.py
```diff
@@ -524,14 +527,35 @@ def _to_disk(
...
+    ipfs_backend: BaseIPFSBackend, prettify: bool, manifest: Manifest
```
pipermerriam: Since the function is named `to_ipfs` maybe drop the prefix on `ipfs_backend ` and just shorten it to `backend`.

ethpm/tools/builder.py
```diff
@@ -524,14 +527,35 @@ def _to_disk(
...
 
     if (cwd / disk_manifest_name).is_file():
         raise ManifestBuildingError(
             "Manifest: {0} already exists in cwd: {1}".format(disk_manifest_name, cwd)
         )
     (cwd / disk_manifest_name).write_text(contents)
     return manifest
+
+
+def to_ipfs(
```
pipermerriam: The name of this is funny to me.  I'd expect it to be `pin_ipfs` or `pin_to_ipfs` as those are still reasonably concise and more explicit about behavior.

ethpm/utils/manifest_validation.py
```diff
@@ -114,3 +115,9 @@ def validate_manifest_exists(manifest_id: str) -> None:
         raise ValidationError(
             "Manifest not found in V2_PACKAGES_DIR with id: {0}".format(manifest_id)
         )
+
+
+def format_manifest(manifest: Manifest, prettify: bool) -> str:
```
pipermerriam: You might change this signature to be `format_manifest(manifest, *, prettify)` so that `prettify` is required to be passed in as a keyword argument.

tests/ethpm/integration/test_ipfs_integration.py
```diff
@@ -26,3 +27,24 @@ def test_local_ipfs_backend_integration_round_trip(monkeypatch):
         local_manifest = f.read()
     ipfs_manifest = backend.fetch_uri_contents(asset_data["Hash"])
     assert ipfs_manifest == local_manifest
+
+
+@pytest.mark.parametrize("backend", (LocalIPFSBackend(), InfuraIPFSBackend()))
```
pipermerriam: While parametrize does a decent job here, using a parametrized fixture would allow you to lazily only instantiate the backend you need and to have better isolation in the case any backend is stateful.

```python
@pytest.fixture(params=[LocalIPFSBackend, InfuraIPFSBackend])
def backend(request):
    return request.param()
```

ethpm/tools/builder.py
njgheorghita: Works for me - I chose `to_ipfs` b/c we have `to_disk` for writing the manifest to disk. Is that a reasonable name or worth changing to something like `write_to_disk`

pipermerriam: I'm a fan of symmetry/consistency for APIs so :+1: .  For some reason `to_disk` is intuitive for me, but `to_ipfs` wasn't

njgheorghita: Sorry is that 👍 to 
1 - `to_disk` and `to_ipfs` 
pr
2 - `pin_to_ipfs` and `write_to_disk`

pipermerriam: 2 (I still think `to_ipfs` is non-intuitive)

PR #111, opened 2018-09-26
add escrow manifest with runtime bytecode to assets

njgheorghita:
### What was wrong?
Need this manifest for testing in `pytest-ethereum`

#### Cute Animal Picture

Cute animal picture


njgheorghita:
merging since minor change

njgheorghita:
@pipermerriam Yup sounds good, all the manifests should be valid a la the json-spec. Are you suggesting just adding a test that validates all the manifests in `ethpm/assets/` - or something different like a separate circleci command that does the validation?

pipermerriam:
Probably just a parametrized test that runs them through the validation, can be part of the existing CI test suite, no need for a new run that I see.

- Make sure the test automatically detects new manifests (no hard coded list)
- Make sure the test sanity checks in case the automated mechanism finds zero manifests which should be an error condition.

PR #114, opened 2018-09-27
Improve error msg on package.get_contract_factory

njgheorghita:
### What was wrong?
I've found the existing error msg to be not so helpful, so added a slight improvement

#### Cute Animal Picture

image


njgheorghita:
merging since minor error msg update

PR #118, opened 2018-10-15
Deprecate 3.5

njgheorghita:
### What was wrong?
Deprecate python 3.5 support, add support for python 3.7

#### Cute Animal Picture
image


pipermerriam:
This should also probably update all of the python 3.5 style comment based type hints.

njgheorghita:
@pipermerriam ping for review - put updating to 3.7 on hold b/c it seems as though it's currently blocked by eth-tester

PR #120, opened 2018-10-24
Add deployments feature to builder tool

njgheorghita:
### What was wrong?
Manifest builder tool needed some functions to facilitate adding a deployment to a manifest

### How was it fixed?
added a `deployment` function, which simply allows you to pass in the desired deployment data

also added a `deployment_type()` function, which is slightly more complex, but allows for managing the same deployment across different chains, and some tx_receipt validation

#### Cute Animal Picture
image


ethpm/tools/builder.py
```diff
@@ -438,12 +441,222 @@ def validate_link_ref(offset: int, length: int, bytecode: str) -> str:
...
+    if tx:
+        yield "transaction", tx
+    if block:
+        yield "block", block
+    if runtime_bytecode:
+        if cytoolz.get_in(
+            ["contract_types", contract_type, "deployment_bytecode", "link_references"],
+            manifest,
+            0,
+        ):
```
njgheorghita: unsure how I feel about auto-populating the linked references from the deployment's corresponding contract type - another option is to simply force the user to provide the linked refs?

docs/tools.rst
```diff
@@ -467,6 +467,106 @@ If you would like to alias your contract type, provide the desired alias as a kw
...
+       ),
+       owned_type(
+           mainnet,
+           address=owned_mainnet_address,
+           transaction=owned_mainnet_transaction,
+           block=owned_mainnet_block,
+           w3=mainnet_w
+       ),
+       escrow_type(
+           testnet,
```
pipermerriam: It is unclear to me from this example what this variable would be.

ethpm/tools/builder.py
```diff
@@ -4,18 +4,21 @@
 from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Tuple
 
 import cytoolz
```
pipermerriam: Looks like you could go ahead and clean up the rest of these to use the `eth_utils.toolz` aliases.

ethpm/tools/builder.py
```diff
@@ -438,12 +441,228 @@ def validate_link_ref(offset: int, length: int, bytecode: str) -> str:
...
+        raise ManifestBuildingError(
+            "A valid w3 instance must be provided to use the deployment_type fn."
+        )
+
+    if not w3.eth.getCode(address):
+        raise ValidationError(f"No code found at the provided address: {address}.")
+        # todo validate code found at address matches runtime_bytecode in contract type.
+
+    if transaction:
+        if not is_hex(transaction):
```
pipermerriam: Note that this will throw a `TypeError` if `transaction` is not a text string.

see: https://github.com/ethereum/eth-utils/issues/137

ethpm/tools/builder.py
```diff
@@ -438,12 +441,228 @@ def validate_link_ref(offset: int, length: int, bytecode: str) -> str:
...
+            raise ValidationError(f"Transaction: {transaction} is not valid hex.")
+
+        tx_receipt = w3.eth.getTransactionReceipt(transaction)
+        if not tx_receipt:
+            raise ValidationError(f"Tx: {transaction} not found on provided w3.")
+
+        if address and tx_receipt.contractAddress != address:
```
pipermerriam: do you want to be forgiving here and use `eth_utils.is_same_address`?

pipermerriam: I also noted that there is a `validate_address` above.  Seems like that validation and this validation should be co-located in the code.

ethpm/tools/builder.py
pipermerriam: If you are unsure about it I would go with the manual solution.  It's harder to remove functionality than it is to add it.

PR #122, opened 2018-10-26
Bugfixes for stabilizing web3 pm-api

njgheorghita:
### What was wrong?
Use assets from `ethpm/assets` rather than `ethpm-spec` submodule
Handle `ipfsapi.ConnectionError` raised when no local node running

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/utils/backend.py
+++ b/ethpm/utils/backend.py
@@ -1,6 +1,8 @@
+import logging
 from typing import Generator, Type
 
 from eth_utils import to_tuple
+from ipfsapi.exceptions import ConnectionError
 
 from ethpm.backends.base import BaseURIBackend
 from ethpm.backends.http import GithubOverHTTPSBackend
@@ -20,6 +22,8 @@ URI_BACKENDS = [
     GithubOverHTTPSBackend,
     RegistryURIBackend,
 ]
+
+logger = logging.getLogger("ethpm.utils.backend")
 
 
 def resolve_uri_contents(uri: str, fingerprint: bool = None) -> bytes:
@@ -53,8 +57,11 @@ def get_translatable_backends_for_uri(
 ) -> Generator[Type[BaseURIBackend], None, None]:
     # type ignored because of conflict with instantiating BaseURIBackend
     for backend in URI_BACKENDS:
-        if backend().can_translate_uri(uri):  # type: ignore
-            yield backend
+        try:
+            if backend().can_translate_uri(uri):  # type: ignore
+                yield backend
+        except ConnectionError:
+            logger.debug("No local IPFS node available on port 5001.", exc_info=True)
 
 
 @to_tuple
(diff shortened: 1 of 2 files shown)


ethpm/utils/backend.py
```diff
@@ -53,8 +54,12 @@ def get_translatable_backends_for_uri(
 ) -> Generator[Type[BaseURIBackend], None, None]:
     # type ignored because of conflict with instantiating BaseURIBackend
     for backend in URI_BACKENDS:
-        if backend().can_translate_uri(uri):  # type: ignore
-            yield backend
+        try:
+            if backend().can_translate_uri(uri):  # type: ignore
+                yield backend
+        # Catch error raised when no local ipfs node running
```
pipermerriam: Emitting a logging message here would probably be a good thing.

ethpm/utils/backend.py
```diff
@@ -53,8 +54,12 @@ def get_translatable_backends_for_uri(
...
+        except ConnectionError:
+            continue
```
pipermerriam: You might emit a logging message here at `DEBUG` level.

ethpm/utils/backend.py
```diff
@@ -65,9 +70,15 @@ def get_resolvable_backends_for_uri(
...
+            if backend_class is default_ipfs:
+                continue
+            # type ignored because of conflict with instantiating BaseURIBackend
+            else:
+                try:
+                    if backend_class().can_resolve_uri(uri):  # type: ignore
+                        yield backend_class
+                # Catch error raised when no local ipfs node running
+                except ConnectionError:
+                    continue
```
pipermerriam: Same here about emitting something at the `DEBUG` log level.

ethpm/utils/backend.py
```diff
@@ -65,9 +72,14 @@ def get_resolvable_backends_for_uri(
...
+        for backend_class in URI_BACKENDS:
+            if backend_class is default_ipfs:
+                continue
+            # type ignored because of conflict with instantiating BaseURIBackend
+            else:
+                try:
+                    if backend_class().can_resolve_uri(uri):  # type: ignore
+                        yield backend_class
+                except ConnectionError:
+                    logger.info("No local IPFS node available on port 5001.")
```
njgheorghita: @pipermerriam I'm not super familiar with the logging module, is this more or less the way to go about it?

pipermerriam:
```python
import logging

logger = logging.getLogger('ethpm.utils.backend')


def my_thing(...):
    logger.info("This is a message that will be logged at info level")
    logger.debug("This one is at debug")
    # https://github.com/ethereum/snake-charmers-tactical-manual/issues/19
    logger.info("We use percent formatting with logging messages: %s", some_arg)
```

That's all you have to do.

pipermerriam: For this specific case you can do:

```python
try:
    ...
except Something:
    logger.debug("Something went wrong", exc_info=true)
```

This will include a full stacktrace in the logging message which can be nice when debugging these things.  Not entirely sure that's the right approach here but probably doesn't hurt.

PR #123, opened 2018-10-27
Update registry to allow 64 byte uris

njgheorghita:
### What was wrong?
Previous vyper registry only allowed URI's up ot 32 bytes - tweaked registry to allow up to 64 byte URIs. Still not ideal since this takes the registry implementation one step further away from the ERC1319 standard, as well as doesn't allow for http uris to be stored on-chain (since they're likely longer than 64 bytes) - but it should accommodate all IPFS uris - which is sufficient for the workshop

#### Cute Animal Picture
image


njgheorghita:
merging since minor tweak to vyper contract - needed for https://github.com/ethereum/web3.py/pull/1125

PR #117, opened 2018-10-13
Add vyper registry and ens manifest data

njgheorghita:
### What was wrong?
Some assets needed for testing in `web3.pm`

#### Cute Animal Picture
image

PR #124, opened 2018-11-15
Cleanup

njgheorghita:
### What was wrong?
re: [discussion](https://github.com/ethpm/py-ethpm/pull/117/files#r233948070)
Turns out the `--exclude  ethpm/assets` isn't even necessary as that folder doesn't have any python files.

#### Cute Animal Picture
image


njgheorghita:
merging since minor cosmetic fixes / assets needed in web3

PR #129, opened 2018-11-29
Update content addressed github uri scheme

### LINKED ISSUE #125: Util for generating valid github uris
### What was wrong?
A util function would be useful that accepts a valid github uri pointing to a manifest json file, and returns the uri as defined [here](https://py-ethpm.readthedocs.io/en/latest/uri_backends.html#http).

### How can it be fixed?
The utility function should be written in https://github.com/ethpm/py-ethpm/blob/master/ethpm/utils/uri.py. It should validate that the given uri leads to a json file hosted on github, [validate](https://github.com/ethpm/py-ethpm/blob/master/ethpm/utils/manifest_validation.py#L59) the json file against the manifest json schema, and return a uri with the specified encoding [here](https://py-ethpm.readthedocs.io/en/latest/uri_backends.html#http). Tests should be written [here](https://github.com/ethpm/py-ethpm/blob/master/tests/ethpm/utils/test_uri_utils.py)

njgheorghita:
### What was wrong?
Decision was made to sunset the fragment based hash verification for github uris and use their native blob system instead - Also mirrors what is suggested in the [ERC 1319](https://github.com/ethereum/EIPs/issues/1319), so this will go towards having a cleaner standard amongst libraries.

### How was it fixed?
Updated the utils and `GithubOverHTTPS` backend to use the blob scheme. 

#### Cute Animal Picture
image

fixes #125


ethpm/backends/http.py
```diff
@@ -1,37 +1,44 @@
...
         response.raise_for_status()
-        validate_uri_contents(response.content, validation_hash)
-        return response.content
+        contents = json.loads(response.content)
+        if contents["encoding"] != "base64":
+            raise CannotHandleURI(
+                "Expected contents returned from Github to be base64 encoded, "
+                f"instead received {contents['encoding']}."
+            )
+        return base64.b64decode(contents["content"])
```
pipermerriam: Maybe I missed it but I expected to see somewhere that we verify that the contents match the hash.  If that isn't present I think it needs to be added.

ethpm/utils/uri.py
```diff
@@ -1,34 +1,73 @@
...
-    authority = parse_result.netloc
-    content_hash = parse_result.fragment
 
-    if not path or not scheme or not content_hash:
+    parsed = parse.urlparse(uri)
+    path, scheme, authority = parsed.path, parsed.scheme, parsed.netloc
+    if not all((path, scheme, authority)):
+        return False
+
+    if [term for term in expected_path_terms if term not in path]:
```
pipermerriam: Generally the pattern for this is to use `any`

```python
if any(term for term in expected_path_terms if not term in path):
    ...
```

The nice part about this is that it is *lazy* meaning it will exit on the first one that is truthy.

ethpm/backends/http.py
njgheorghita: Yup, I chose to remove that for some reason that escapes me now, I'll set about re-implementing that.

njgheorghita:
@pipermerriam ping - added a validation util to verify that the hash of the contents retrieved from a github blob uri match the hash in the uri

PR #131, opened 2018-12-03
Update builder to allow for missing linkreferences in compiler output

njgheorghita:
### What was wrong?
Builder's strategy to check for link-refs in compiler output was poor.

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/tools/builder.py
+++ b/ethpm/tools/builder.py
@@ -373,13 +373,13 @@ def normalize_contract_type(
 
 @to_dict
 def normalize_bytecode_object(obj: Dict[str, Any]) -> Iterable[Tuple[str, Any]]:
-    if obj["linkReferences"]:
-        yield "link_references", process_link_references(
-            obj["linkReferences"], obj["object"]
-        )
-        yield "bytecode", process_bytecode(obj["linkReferences"], obj["object"])
+    link_references = obj.get("linkReferences")
+    bytecode = obj.get("object")
+    if link_references:
+        yield "link_references", process_link_references(link_references, bytecode)
+        yield "bytecode", process_bytecode(link_references, bytecode)
     else:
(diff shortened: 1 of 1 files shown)


njgheorghita:
merging since minor bugfix

ethpm/tools/builder.py
```diff
@@ -373,13 +373,13 @@ def normalize_contract_type(
 
 @to_dict
 def normalize_bytecode_object(obj: Dict[str, Any]) -> Iterable[Tuple[str, Any]]:
-    if obj["linkReferences"]:
-        yield "link_references", process_link_references(
-            obj["linkReferences"], obj["object"]
-        )
-        yield "bytecode", process_bytecode(obj["linkReferences"], obj["object"])
+    link_references = obj.get("linkReferences")
```
pipermerriam: We try to avoid return value checking due to it being error prone (it is easy to forget to make the check).  Maybe use a `try/except` pattern instead?

ethpm/tools/builder.py
```diff
@@ -373,13 +373,13 @@ def normalize_contract_type(
...
+    bytecode = obj.get("object")
+    if link_references:
+        yield "link_references", process_link_references(link_references, bytecode)
```
pipermerriam: it's not clear to me that all of the underlying functions will properly handle `bytecode` being `None` (which is now possible)

ethpm/tools/builder.py
```diff
@@ -373,13 +373,13 @@ def normalize_contract_type(
...
+        yield "bytecode", process_bytecode(link_references, bytecode)
     else:
-        yield "bytecode", add_0x_prefix(obj["object"])
+        yield "bytecode", add_0x_prefix(bytecode)
```
pipermerriam: I'm pretty sure this will fail if `bytecode is None`

PR #132, opened 2018-12-10
Bugfix the bugfix

njgheorghita:
### What was wrong?
Bugs in #131 

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/tools/builder.py
+++ b/ethpm/tools/builder.py
@@ -385,8 +385,17 @@ def normalize_contract_type(
 
 @to_dict
 def normalize_bytecode_object(obj: Dict[str, Any]) -> Iterable[Tuple[str, Any]]:
-    link_references = obj.get("linkReferences")
-    bytecode = obj.get("object")
+    try:
+        link_references = obj["linkReferences"]
+    except KeyError:
+        link_references = None
+    try:
+        bytecode = obj["object"]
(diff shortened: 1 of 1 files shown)


ethpm/tools/builder.py
```diff
@@ -385,8 +385,11 @@ def normalize_contract_type(
 
 @to_dict
 def normalize_bytecode_object(obj: Dict[str, Any]) -> Iterable[Tuple[str, Any]]:
-    link_references = obj.get("linkReferences")
-    bytecode = obj.get("object")
+    try:
+        link_references = obj["linkReferences"]
+    except KeyError:
+        link_references = None
+    bytecode = obj["object"]
```
pipermerriam: What about this?  How do we know that `obj` has an `"object"` key?

njgheorghita: the `"object"` key in `obj` has the actual bytecode as its value in the solc output, so it should always be present if the user is passing in valid solc output (whereas `link_references` is optional), which is why I didn't think it was necessary to catch any errors. Though, I can add another try/except if you think it's useful?

pipermerriam: users will eventually make every mistake in the book including this one so I think adding a good error message now saves the time later to deal with the github issue about the "bug".

PR #133, opened 2018-12-10
Ensure all source paths created by builder are relative

njgheorghita:
### What was wrong?
Builder was creating invalid paths for "sources", unless path was relative in the compiler output, it wouldn't be added to the manifest as a relative path, which is required by the [spec](http://ethpm.github.io/ethpm-spec/package-spec.html#format)

### How was it fixed?
Added a util to check if path is relative - and return properly formatted path

#### Cute Animal Picture
image


ethpm/tools/builder.py
```diff
@@ -134,12 +134,22 @@ def get_names_and_paths(compiler_output: Dict[str, Any]) -> Dict[str, str]:
...
         for contract_name in compiler_output[path].keys()
     }
 
 
+def make_path_relative(path: str) -> str:
+    """
+    Returns the given path prefixed with "./" if the path
+    is not already relative in the compiler output.
+    """
+    if path[:2] != "./":
```
pipermerriam: I can't remember the state of things so this is somewhat homework for you to do.  Can you double-check/verify that we aren't subject to other types of incidental invalid paths being added.  For example, if the path string was `../../Owned.sol` this would produce `./../../Owned.sol` which *should* be invalid since it traverses outside of the *virtual source tree* for this manifest's contracts.

If this is already in place just ignore me my paranoia

njgheorghita: @pipermerriam Sure, i'll definitely look into it. Can you expand a little on *virtual source tree* - what's inside / outside of the tree?

pipermerriam:
So a manifest has a set of source files, all of which exist in the *virtual source tree*.

so a file with the path `./owned.sol` would exist in the root directory of that *virtual source tree*.  The concept of installation would involve writing that source tree to disk somewhere and ensuring that none of the paths traverse outside of the root of the virtual source tree is how we can ensure that packages which *would* deposit files outside of that directory would be considered invalid.

PR #134, opened 2018-12-11
Add compiler output and natspec to manifest builder

njgheorghita:
### What was wrong?
Manifest builder wasn't capable of adding natspec and compiler info to a contract type.

#### Cute Animal Picture
image


ethpm/tools/builder.py
```diff
@@ -341,6 +342,25 @@ def _contract_type(
...
+    Raises exception if selected field data is not available in the contract type data
+    automatically gathered by normalize_compiler_output. Otherwise, returns the data.
+    """
+    for field in selected_fields:
+        if field in all_type_data:
+            yield field, all_type_data[field]
+        else:
+            raise ManifestBuildingError(
+                f"Selected field: {field} not available in data collected from solc output: "
+                f"{list(all_type_data.keys())}. Please make sure the relevant data is present "
```
pipermerriam: *nitpick*

since dictionary keys aren't ordered this can produce two different messages with the same data.  Normally when I dump unordered data into a logging/user message I first sort it.

ethpm/tools/builder.py
```diff
@@ -379,8 +399,27 @@ def normalize_contract_type(
             yield "runtime_bytecode", normalize_bytecode_object(
                 contract_type_data["evm"]["deployedBytecode"]
             )
-    # todo support natspec.
-    yield "natspec", {}
+    if any(key in contract_type_data for key in ["devdoc", "userdoc"]):
```
pipermerriam: *nitpick*

1. When doing checks for inclusion you should use a set `{"devdoc", "userdoc"}` as they are better suited for inclusion checks.
2. The value should probably be elevated to a top level constant `NATSPEC_FIELDS = {'devdoc', 'userdoc'}`

ethpm/tools/builder.py
```diff
@@ -379,8 +399,27 @@ def normalize_contract_type(
...
+
+
+@to_dict
+def normalize_compiler_object(obj: Dict[str, Any]) -> Iterable[Tuple[str, Any]]:
+    yield "name", "solc"
+    yield "version", obj["compiler"]["version"]
+    yield "settings", {"optimize": obj["settings"]["optimizer"]["enabled"]}
+
+
+def normalize_natspec_object(obj: Dict[str, Any]) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
pipermerriam: I don't recall exactly why but I think you need to do a *deep* merge:

https://github.com/ethereum/populus/blob/4e9cc1efcea2333f48ceec9081ef5392d87608e8/populus/packages/build.py#L199-L204

PR #137, opened 2018-12-15
Update to web3 v5

njgheorghita:
### What was wrong?
update to web3 v5 and remove eth-abi dependency

#### Cute Animal Picture
🦔


mikeshultz:
Would be pretty cool to get the `web3[tester]` splinter pulled out, or optional somehow.  The `web3[tester]` splinter/extras don't look like they're required to run py-ethpm as a user.  Not sure if this is the best way to go about it, but maybe something like this? 

    from setuptools.command.develop import develop as _develop
    from subprocess import check_call

    [...]

    class develop(_develop):
        """ Post-installation to make sure web3 tester extras are installed."""
        def run(self):
            _develop.run(self)
            check_call(['pip', 'install', 'web3[tester]>=5.0.0a2,<6.0.0'])

    setup(
        [...]
        install_requires=[
            [...]
            'web3>=5.0.0a2,<6.0.0',
        ],
        [...]
        cmdclass={
            'develop': develop,
        },
    )

setup.py
```diff
@@ -54,16 +54,17 @@
...
         'ipfsapi>=0.4.3,<1',
         'jsonschema>=2.6.0,<3',
         'protobuf>=3.0.0,<4',
-        'pytest-ethereum>=0.1.3a.5,<1',
+        'pytest-ethereum>=0.1.3a.6,<1',
+        'py-evm>=0.2.0a38,<1',
         'py-solc>=3.2.0,<4',
         'rlp>=1.0.1,<2',
-        'web3[tester]>=4.7,<5',
+        'web3[tester]>=5.0.0a2,<6.0.0',
```
pipermerriam: Is this a testing dependency?

njgheorghita: @pipermerriam The `[tester]` splinter is a testing dependency, but `web3` by itself is needed for a user to use `py-ethpm`. 

I tried moving `web3[tester]>=5.0.0a3,<6.0.0` to the `test` section of `extras_require` and leaving `web3>=5.0.0a3,<6.0.0` in the `install_requires` section, in the hopes that the `[tester]` splinter would only be installed with `[test]` or `[dev]`. But this failed to install `eth_tester` when running `pip install -e .[dev]`. 

Is there a better way to handle only installing the `[tester]` splinter in a the `extras_require[test]` env, but the same `web3` version without the `[tester]` splinter in `install_requires`?

njgheorghita:
@mikeshultz Noted - created issue #140 to track, but moving ahead to merge this

PR #139, opened 2018-12-20
Update set_default_w3 to return a new Package instance

njgheorghita:
### What was wrong?
`Package.set_default_w3` was stateful and confusing to manage, and I found it more useful to simply return a new package instance with the updated w3 rather than worrying about managing the state (eg. to which chain is the current w3 instance connected) of a `Package` instance.

#### Cute Animal Picture
image

### DIFF
--- a/tests/ethpm/test_package.py
+++ b/tests/ethpm/test_package.py
@@ -2,7 +2,7 @@ from eth_utils import is_same_address, to_canonical_address
 import pytest
 from web3 import Web3
 
-from ethpm.exceptions import InsufficientAssetsError
+from ethpm.exceptions import InsufficientAssetsError, ValidationError
 from ethpm.package import Package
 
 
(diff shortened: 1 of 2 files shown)


njgheorghita:
@pipermerriam ping!

tests/ethpm/test_package.py
```diff
@@ -21,16 +21,19 @@ def deployed_safe_math(safe_math_package, w3):
...
     new_w3 = Web3(Web3.EthereumTesterProvider())
-    current_package = Package(all_manifests, w3)
-    assert current_package.w3 is w3
-    current_package.set_default_w3(new_w3)
-    assert current_package.w3 is new_w3
+    original_package, _ = deployed_safe_math
+    assert original_package.w3 is w3
+    new_package = original_package.set_default_w3(new_w3)
+    assert new_package.w3 is new_w3
+    assert original_package.manifest == new_package.manifest
```
pipermerriam: You can do one more check of `assert new_package is not original_package` since they should be distinct objects.

PR #142, opened 2019-01-17
Add global setup for doctest

njgheorghita:
### What was wrong?
Some of the autodoc examples were failing b/c of undefined variables. It seems to me as though `testsetup::` doctest directive won't work for autodoc, and the only workaround is to use `doctest_global_setup` a la [here](https://stackoverflow.com/questions/27231362/sphinx-doctest-vs-autodoc)

#### Cute Animal Picture
image


njgheorghita:
Merging since relatively minor & needed for release

PR #145, opened 2019-02-12
Update vyper registry contracts

njgheorghita:
### What was wrong?
- Old registry reference vyper implementations broke with new vyper version
- Event missing that was recently added to ERC 1319

### How was it fixed?
- Fixed vyper and added event

#### Cute Animal Picture
image


njgheorghita:
merging since it's simply updating some assets

pipermerriam:
Breaking this contract out into it's own repo and using twig for continued development (at some point) will be a good dogfooding excercise.  Especially because we have the registry that is behind the `snakecharmers.eth` ENS name and with this update, you'll also need to 1) deploy the new contract and 2) migrate the data from the previous one to the new one which excercises the migration story for a twig project.

Issue #148, opened 2019-03-21
Infura Problems with Required API key


kclowes:
### What was wrong?
If there is no Infura environment variable set, web3 v5 tests can't run because ethpm is loading Infura here: https://github.com/ethpm/py-ethpm/blob/master/ethpm/backends/registry.py#L5

### How can it be fixed?
@pipermerriam suggested a few things:
Since we need to get this out before 3/27, we can back out ethpm from v5 for a minute, and then add it back in after we've made "the fix"
"The fix" might look like: 
- Passing a web3 instance to ethpm (least user friendly option, but least error prone)
- Check for the environment variable and auto connect to infura if the environment variable is set
- Something else...?

Probably all of these options require some documentation updates

njgheorghita:
piggy backing off this issue 

> Removing it from constants and coming up with the mechanism that 1) doesn't require users to use infura since it effectively leaks privacy information and we shouldn't have a hard dependency on a centralized solution but 2) still preserves as much ease of usability as possible.

I recognize moving away from infura probably goes opposite of conversations you and I had about it being a fine solution, but with this stuff moving into production I think it's now time to address that particular piece of debt.

This probably really means leaving the ethpm library such that it require more manual configuration and addressing the usability component in the web3.py codebase.

PR #149, opened 2019-03-22
Load infura url manually

njgheorghita:
### What was wrong?
#148 

### How was it fixed?
- removed automatically created infura `w3` instance in favor of manually generated `w3` endpoint using infura token

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/backends/registry.py
+++ b/ethpm/backends/registry.py
@@ -2,7 +2,8 @@ import os
 from urllib import parse
 
 from eth_utils import to_bytes
-from web3.auto.infura import w3
+from web3 import Web3
+from web3.providers.auto import load_provider_from_uri
 
 from ethpm.backends.base import BaseURIBackend
 from ethpm.constants import INFURA_API_KEY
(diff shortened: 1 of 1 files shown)


njgheorghita:
@kclowes Let me know if this is a good solution that you had in mind. If so, i'll merge and cut a new release today

ethpm/backends/registry.py
```diff
@@ -2,7 +2,8 @@
 from urllib import parse
 
 from eth_utils import to_bytes
-from web3.auto.infura import w3
+from web3 import Web3
+from web3.providers.auto import load_provider_from_uri
 
 from ethpm.backends.base import BaseURIBackend
 from ethpm.constants import INFURA_API_KEY
```
pipermerriam: Can we get a tracking issue for this?  We should probably plan to be rid of it within a month or two at the very most.

njgheorghita: Just in terms of removing the `INFURA_API_KEY` from `constants.py` and requiring users to provide their own?

pipermerriam: Removing it from constants and coming up with the mechanism that 1) doesn't require users to use infura but 2) still preserves as much ease of usability as possible.  

This probably really means leaving the `ethpm` library such that it require *more* manual configuration and addressing the usability component in the web3.py codebase.

kclowes:
I think this should work @njgheorghita. Thank you!

PR #151, opened 2019-04-05
Update registry backend to work with web3.pm

njgheorghita:
### What was wrong?
`RegistryURIBackend` was written before `web3.pm` so it's been updated to work with the new api.

Wrote a `parse_registry_util` since it's now needed in multiple places

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/utils/uri.py
+++ b/ethpm/utils/uri.py
@@ -1,3 +1,4 @@
+from collections import namedtuple
 import hashlib
 import json
 from typing import Tuple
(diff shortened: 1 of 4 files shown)


ethpm/utils/uri.py
```diff
@@ -90,3 +91,15 @@ def validate_blob_uri_contents(contents: bytes, blob_uri: str) -> None:
...
+def parse_registry_uri(uri: str) -> Tuple[str, ...]:
+    """
+    Validate and return (authority, pkg name, version) from a valid registry URI
+    """
+    validate_registry_uri(uri)
+    parsed_uri = parse.urlparse(uri)
+    authority = parsed_uri.netloc
+    pkg_name = parsed_uri.path.strip("/")
+    pkg_version = parsed_uri.query.lstrip("version=").strip("/")
+    return authority, pkg_name, pkg_version
```
pipermerriam: Maybe use a `NamedTuple` for the return type here?  That's probably ideal so that it's easy for consumers of this API to keep this info packaged up together while maintaining strong type safety and having property access to the individual components.  (make sure the class is available in the public namespace so that consumers of this package can use it in their type hints)

PR #152, opened 2019-04-23
Write is_supported_content_addressed_uri

njgheorghita:
### What was wrong?
Already used this function in both `ethpm-cli` and the ipfs pinner, so it seems like it belongs here.

#### Cute Animal Picture
image

### DIFF
--- a/ethpm/__init__.py
+++ b/ethpm/__init__.py
@@ -1,7 +1,6 @@
+from pathlib import Path
 import sys
 import warnings
-
-from pathlib import Path
 
 
 if sys.version_info.major < 3:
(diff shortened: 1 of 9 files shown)


njgheorghita:
@pipermerriam What exactly do you have in mind by '3rd party URIs'? Can you give an example? I was under the impression that it was pretty much IPFS, Github blob, Swarm (eventually) that will be supported/valid content-addressed uris.

pipermerriam:
@njgheorghita bitbucket, gitlab, gists, .... ipfs-gateway uris (there can be more than one gateway, cloudflair has one).  The EthPM spec is intentionally open about the definition of a content addressable URI and thus, there are tons of options out there for people to include the content hash in different URI schemes.

PR #153, opened 2019-04-25
Update mypy

njgheorghita:
### What was wrong?
New mypy version has been released, was updated along with other linting dependencies

#### Cute Animal Picture
image


ethpm/contract.py
```diff
@@ -90,13 +90,9 @@ def validate_attr_dict(self, attr_dict: Dict[str, str]) -> None:
...
             )
 
-        all_link_refs: Tuple[Any, ...]
-        if self.unlinked_references and self.linked_references:
-            all_link_refs = self.unlinked_references + self.linked_references
-        elif not self.unlinked_references:
-            all_link_refs = self.linked_references
-        else:
-            all_link_refs = self.unlinked_references
+        unlinked_refs = self.unlinked_references if self.unlinked_references else ({},)
```
pipermerriam: This can be shortened to `unliked_refs = self.unlinked_references or ({},)

ethpm/contract.py
```diff
@@ -15,11 +15,11 @@ class LinkableContract(Contract):
     contract factories with link references in their package's manifest.
     """
 
-    unlinked_references: Tuple[Dict[str, Any]] = None
-    linked_references: Tuple[Dict[str, Any]] = None
+    unlinked_references: Optional[Tuple[Dict[str, Any]]] = None
```
pipermerriam: Should the inner dict be mutable?  If not maybe convert to `FrozenDict` at which point you could assign `(frozendict(),)` as a class property since it's immutable and then not have to worry about the setting of the default value in the constructor.

ethpm/deployments.py
```diff
@@ -32,7 +32,8 @@ def __contains__(self, key: str) -> bool:
 
     def get(self, key: str) -> Dict[str, str]:
         self._validate_name_and_references(key)
-        return self.deployment_data.get(key)
+        # type ignored b/c _validate_name_and_references enforces key exists
+        return self.deployment_data.get(key)  # type: ignore
```
pipermerriam: With the comment above why not use `return self.deployment_data[key]` since we know the key exists?

ethpm/package.py
```diff
@@ -130,7 +132,7 @@ def manifest_version(self) -> str:
         return self.manifest["manifest_version"]
 
     @property
-    def uri(self) -> str:
+    def uri(self) -> Optional[str]:
```
pipermerriam: Why `Optional`?  Assuming this is to take into account packages initialized without a `URI`, what if we just generated a [data URI](https://en.wikipedia.org/wiki/Data_URI_scheme) for those?  Data URIs could **always** be valid since they contain their content.

ethpm/utils/manifest_validation.py
```diff
@@ -112,7 +112,7 @@ def validate_manifest_exists(manifest_id: str) -> None:
         )
 
 
-def format_manifest(manifest: Manifest, *, prettify: bool) -> str:
+def format_manifest(manifest: Manifest, *, prettify: Optional[bool]) -> str:
```
pipermerriam: I think you want to do `prettify: bool = None` which `mypy` will interpret as `Optional`.  Otherwise the function signature doesn't really make sense since at the python level you still are required to provide the argument *but* it is allowed to be `None` (which is weird)

ethpm/backends/ipfs.py
```diff
@@ -180,5 +180,6 @@ def get_ipfs_backend(import_path: str = None) -> BaseIPFSBackend:
 
 def get_ipfs_backend_class(import_path: str = None) -> Type[BaseIPFSBackend]:
     if import_path is None:
-        import_path = os.environ.get("ETHPM_IPFS_BACKEND_CLASS", DEFAULT_IPFS_BACKEND)
+        backend_cls = os.environ.get("ETHPM_IPFS_BACKEND_CLASS", DEFAULT_IPFS_BACKEND)
+        return import_string(backend_cls)
```
pipermerriam: Why was this needed?

PR #156, opened 2019-05-07
Refactor utils/ to _utils/

njgheorghita:
### What was wrong?
`utils/` folder should be `_utils` to imply private api-ness
`ethpm.typing` is no longer necessary with new `eth_typing` version

re [comment here](https://github.com/ethpm/ethpm-cli/pull/3#discussion_r281313429)

#### Cute Animal Picture
image


njgheorghita:
This is a fairly breaking change, so it's more or less on hold until the whole chain of dependencies is compatible

njgheorghita:
closing since refactor is happening in #163

PR #159, opened 2019-05-23
Fix link refs bug and update dependencies

njgheorghita:
### What was wrong?
Bugfix on how linkrefs are processed in `LinkableContract`.
Updated some dependencies.

#### Cute Animal Picture
image


njgheorghita:
merging since minor & bugfix is needed in new release

Issue #140, opened 2019-01-16
Move `web3[tester]` splinter to `extras_require[test]`


njgheorghita:
See the conversation [here](https://github.com/ethpm/py-ethpm/pull/137)

schmir-at-bb:
If someone install web3 version 5 with `pip install web3==5.0.0b2`, he ends up with `ethpm==0.1.4a19`.

Requiring web3 with pkg_resources fails:
```
..stlines/protocol/contract-deploy-tools (git)-[master] % python
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pkg_resources; pkg_resources.require("web3")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/bbralf/trustlines/venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 900, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/home/bbralf/trustlines/venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 786, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'pytest-ethereum<1.0.0,>=0.1.3a6; extra == "tester"' distribution was not found and is required by web3
>>> 
```

Removing `[tester]` from the `Requires-Dist: web3[tester] (<6,>=5.0.0b1)` line  in `.../site-packages/ethpm-0.1.4a19.dist-info/METADATA` makes the above `pkg_resources.require` succeed.

I do think this is a serious issue that needs to be fixed. It basically breaks every script using setuptools console_scripts with web3 as a dependency.

PR #160, opened 2019-06-12
Remove web3[tester] dep from core dependencies

njgheorghita:
### What was wrong?
#140 // https://github.com/ethpm/py-ethpm/pull/137#issuecomment-449873953

Core dependency of `web3[tester]` causes when doing a base install.

Thanks @mikeshultz for the solution!

#### Cute Animal Picture
image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,81 +2,91 @@
 # -*- coding: utf-8 -*-
 import os
 
-from setuptools import (
-    setup,
-    find_packages,
-)
+from setuptools.command.develop import develop as _develop
+from setuptools import setup, find_packages
+from subprocess import check_call
 
 
 DIR = os.path.dirname(os.path.abspath(__file__))
+readme = open(os.path.join(DIR, "README.md")).read()
+WEB3_VERSION = "5.0.0b2,<6.0.0"
 
 
-readme = open(os.path.join(DIR, 'README.md')).read()
+class develop(_develop):
+    """
+    Post-installation to make sure web3 tester extras are installed.
+    """
 
-extras_require={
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -2,81 +2,91 @@
...
+
+
+class develop(_develop):
+    """
+    Post-installation to make sure web3 tester extras are installed.
+    """
 
+    def run(self):
+        _develop.run(self)
+        check_call(["pip", "install", f"web3[tester]>={WEB3_VERSION}"])
```
pipermerriam: I don't think this will work when it's installed via a wheel (but I'm not positive) you should confirm before proceeding.  Generally this is in the realm of things that I consider bad ideas that don't end up reliably working beyond the development environment.

mikeshultz: I agree it's a little iffy, but isn't this a command that will *only* be run in a development/testing environment?

njgheorghita: > I don't think this will work when it's installed via a wheel (but I'm not positive) 
This is true from what I've read.

> isn't this a command that will only be run in a development/testing environment?
I feel the same way, but I guess it's possible other projects might have a core dependency on `ethpm[dev]` - which would be weird, but possible. IMO though, this is the cleanest solution I've come across for removing the `web3[tester]` splinter from ethpm's core dependencies - and worth the trade-off, but I could be misjudging this.

pipermerriam:
given this is a dependency of web3.py I'm nervious, but I don't have a specific failure in mind, only the generic idea that this misbehaves in some un-expected way.  Thoughts on just documenting the appropriate installation steps well?

PR #163, opened 2019-06-24
Move utils to _utils namespace

njgheorghita:
### What was wrong?

WIP

`/utils` folder needs moving to `/_utils` to make the web3 merge cleaner and more consistent with web3 library.

**THIS IS ONLY ABOUT HALF OF THE TOTAL REFACTOR**
I wanted to submit a pr to get feedback on whether or not this is moving in the right direction. As I understand it, anything in `/_utils` should NOT be part of the general API and only used internally.

#### Cute Animal Picture
image


ethpm/_utils/backend.py
```diff
@@ -13,51 +13,25 @@
     get_ipfs_backend_class,
 )
 from ethpm.backends.registry import RegistryURIBackend
-from ethpm.exceptions import CannotHandleURI
 from ethpm.typing import URI
 
-URI_BACKENDS = [
+logger = logging.getLogger("ethpm.utils.backend")
+
+ALL_URI_BACKENDS = [
```
pipermerriam: orthogonal but probably worth adopting a registry pattern similar to what was done here: https://github.com/ethereum/trinity/pull/586/files#diff-bd2ae038cb9e0fb653d90082dcf5ddfcR42

ethpm/_utils/chains.py
```diff
@@ -0,0 +1,102 @@
...
+from typing import Tuple
+from urllib import parse
+
+from eth_utils import add_0x_prefix, remove_0x_prefix, to_hex
+from web3 import Web3
+
+from ethpm.typing import URI
+
+
+def get_genesis_block_hash(web3: Web3) -> str:
```
pipermerriam: @kclowes maybe this makes sense to be a new API on the `w3.eth` module.  Something like `w3.eth.get_genesis_hash()`

ethpm/_utils/chains.py
```diff
@@ -0,0 +1,102 @@
...
+    See: https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki
+    """
+    if resource_type != BLOCK:
+        raise ValueError("Invalid resource_type.  Must be one of 'block'")
+    elif not is_block_or_transaction_hash(resource_identifier):
+        raise ValueError(
+            "Invalid resource_identifier.  Must be a hex encoded 32 byte value"
+        )
+    elif not is_block_or_transaction_hash(chain_id):
+        raise ValueError("Invalid chain_id.  Must be a hex encoded 32 byte value")
```
pipermerriam: are these functions all just copy/pasted or do they need to be reviewed?

njgheorghita: everything's copy/pasted as far as what's inside a function body

njgheorghita:
@pipermerriam What's the best strategy to get these breaking changes merged and compatible with pytest-ethereum/web3 without breaking everything? I was planning to minor version bump ethpm from `v0.1.4a19` to `v0.1.5a1`. In which case all that needs to happen is pte and web3 need new versions released which raise their dependency to `ethpm>=0.1.5a1`? Does that sound about right?

njgheorghita:
Another option, since this codebase is heading towards web3 anyway, is to just not cut another release of ethpm and migrate the code over to web3?

PR #164, opened 2019-06-25
Implement write_to_disk for ipfs backends

njgheorghita:
### What was wrong?
This is something I find myself doing a lot manually, and would be useful if available on uri backends.

#### Cute Animal Picture
image


njgheorghita:
Yeah, that's a fair point. I'm not against moving this logic to a util function. I was just thinking it might be cleaner/simpler since you already have the IPFS backend to do something like 
```python
ipfs_backend.write_to_disk(uri, target_path)
```
rather than
```python
import write_uri_to_disk
write_uri_to_disk(ipfs_backend, uri, target_path)
```
In terms of duplicating the logic, I just moved the definition of `write_to_disk` to the `BaseURIBackend`, but if you think this is still better off as a utility function, I can see that point of view.

njgheorghita:
I guess this relates to how we interact with the ipfs backend in general. Typically, I find myself instantiating a specific backend (i.e. `InfuraIPFSBackend` / `LocalIPFSBackend`) and using them directly. But with the current update to `resolve_uri_contents(uri) -> contents` which automatically tries all the backends to fetch the contents, it might make sense to have a similar function `write_uri_to_disk(uri, target_path)` which is backend agnostic and just tries to fetch from whatever backend is able/available. Encouraging users to skip instantiating their own URI backend to fetch contents and use the utils `resolve_uri_contents()` / `write_uri_to_disk`.

PR #167, opened 2019-07-03
Write build dependency mechanism for manifest building

njgheorghita:
### What was wrong?
`builder` tool needs a mechanism for adding build dependencies to a manifest.

#### Cute Animal Picture
image

### DIFF
--- a/tests/ethpm/tools/test_builder.py
+++ b/tests/ethpm/tools/test_builder.py
@@ -664,27 +664,27 @@ def test_builder_deployment_type_complex(escrow_package):
     assert len(list(manifest["deployments"].values())[1]) == 2
 
 
-def test_builder_with_build_dependencies():
-    expected_single_build_dep = {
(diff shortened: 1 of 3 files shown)


tests/ethpm/tools/test_builder.py
```diff
@@ -661,3 +662,49 @@ def test_builder_deployment_type_complex(escrow_package):
...
+        BASE_MANIFEST,
+        build_dependency(
+            "package", "ipfs://QmUYcVzTfSwJoigggMxeo2g5STWAgJdisQsqcXHws7b1FW"
+        ),
+        build_dependency(
+            "escrow", "ipfs://QmPDwMHk8e1aMEZg3iKsUiPSkhHkywpGB3KHKM52RtGrkv"
+        ),
+    )
+    assert actual_single == expected_single
+    assert actual_double == expected_double
```
pipermerriam: This looks like it should be split into two tests.

PR #169, opened 2019-07-11
Update registry uri

njgheorghita:
### What was wrong?
For the registry uri to be useful, it needs an idea of what chain the registry lives on, so I added the chain id to the registry uri definition.
```
erc1319://address:chain_id/package_name?version=1.0.0
```
IMO this will help out a lot with simplifying some of the ethpm-cli commands, since we want to support registries on testnets, and now the user only has to provide the uri to select the correct registry.

NOTE: This pr is located here for now, but will be moved to web3 asap

#### Cute Animal Picture
image


njgheorghita:
closing since pr has been moved to `web3/ethpm` https://github.com/ethereum/web3.py/pull/1382

